PR #5, opened 2020-01-28
Broadcast

meggart:


Project.toml
```diff
@@ -13,4 +13,4 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 test = ["Test","Statistics"]
 
 [compat]
-julia = "^1"
+julia = "1.2"
```
visr: For my understanding, why does this package require at least 1.2?

meggart: It is to allow lazy broadcasting. There were some changes in the internal broadcasting machinery that are only available in Julia 1.2

Do you think it would be worth the effort to allow check the Julia version and have at least efficient getindex/setindex? There is also a failing test for reductions on julia 1.0 but I would know a workaround for this one.

visr: Ok I see. I wouldn't put in that effort now. Perhaps if some package wants to use it but really needs julia 1.0 support, but I'm not aware of any.

PR #31, opened 2021-05-19
fix common_chunks for last axis length 1

rafaqz:
Seems that this was only broken when the final axis length is 1 for one of the arrays, so I added a new test block to catch this specifically.


meggart (approved):
Thanks a lot

PR #36, opened 2021-06-06
fix ReshapeDiskArray and PermutDiskArray constructors

rafaqz:
This is a tiny PR just to fix the behavior of these constructors. Currently they can't be reconstructed from their fields, which makes them hard to reconstruct with new contents.


meggart (approved):

meggart:
Thanks for the fix, I just registered a new version.

Issue #37, opened 2021-06-19
Implement chunked `foreach`/non allocating chunked iteration.


rafaqz:
As pointed out in #15 broadcast can do pretty much everything.  But one issue is you have to call `collect` and allocate an array?

In GeoData.jl I'm using broadcast to iterate over arrays to e.g. calculate ranges of non nodata values for a `trim` method, using `broadcast` so the operation is chunked. But I still need to call `collect` on the broadcast object, which allocates.

It would be good if there was a way to iterate over the object with chunking that didn't allocate. Maybe that's possible already?

rafaqz:
Maybe we can achieve this in a more general way by using `Iterators.Stateful` to iterate over the chunks, caching the chunks along the column as we load them and swapping them out as required. 

On the first iteration we can allocate them required memory for the number of chunks along the column, and copy to it from disk when the iterator gets to the next chunk.

meggart:
I am still not sure I completely understand the issue here. What exactly is the use case for iteration here that can not be accomplished using `reduce` or `reducedim`? You can also do reductions over broadcasted objects, so it should not be necessary to allocate the full array.

PR #39, opened 2021-06-25
fix broadcast for iterables like Ref and Tuple

rafaqz:
Currently broadcast with `Ref` or `Tuple` does not work.

The fix is trivial: to free the type in `subsetarg(x::Number, x) = x` from `Number` to `Any`. I'm not totally accross what `subsetarg` is for but this seems to work.

### DIFF
--- a/src/ops.jl
+++ b/src/ops.jl
@@ -60,7 +60,7 @@ function common_chunks(s,args...)
     return map(i->i[1],tt),map(i->i[2],tt)
   end
 end
-subsetarg(x::Number,a) = x
+subsetarg(x, a) = x
 function subsetarg(x::AbstractArray,a)
   ashort = maybeonerange(size(x),a)
   view(x,ashort...) #Maybe making a copy here would be faster, need to check...
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -121,7 +121,7 @@ function test_broadcast(a_disk1)
   a_disk2 = _DiskArray(rand(1:10,1,9), chunksize=(1,3))
   a_mem   = reshape(1:2,1,1,2);
 
-  s = a_disk1 .+ a_disk2
+  s = a_disk1 .+ a_disk2 .* Ref(2) ./ (2,)
   #Test lazy broadcasting
   @test s isa DiskArrays.BroadcastDiskArray
   @test getindex_count(a_disk1)==0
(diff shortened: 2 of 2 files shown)


meggart:
Thanks for catching the bug, but this fix does not yet work for all cases. Consider this test case:

````julia
using Zarr
z = zzeros(Int, 10, chunks=(5,))
out = zeros(Int, 10)
out .= z .+ (1:10...,)
````
still errors on this branch. Basically what `subsetarg` does is to subset the parent arrays of the broadcast expression to the current chunk while respecting broadcast rules for singleton dimensions. So the following works for Tuples and Refs:

````julia
subsetarg(x, a) = x
subsetarg(x::Tuple, a) = x[a...]
subsetarg(x::Ref, a) = x[]
function subsetarg(x::AbstractArray,a)
  ashort = maybeonerange(size(x),a)
  view(x,ashort...) #Maybe making a copy here would be faster, need to check...
end
````
Would you want to update the PR or shall I do it? I think we should add a test similar to the one above as well.

rafaqz:
Oh right. Im short on time for the next month so if you can make the change go for it. 

Im also wondering if it will work for all other iterables?

meggart:
Ok I will update this PR. Supporting iterables that don't define `getindex` might be tough, because you have to be able to random-access a certain chunk/block of the data. I can make the method for Tuples the default fallback so that everything which has `getindex` defined should work, but probably not all iterators.

rafaqz:
Oh right that makes sense. Very hard to iterate with chunking.

This could be another example where my cached `iterate` strategy from #37 could work... if we cache chunks down the column we can iterate linearly without getindex, and just keep reusing the same cache memory as we move across columns. 

It would use more memory that the current strategy but still a lot less than the whole array.

rafaqz:
@meggart how do you feel about merging this as-is? It would at least mean that I can merge https://github.com/rafaqz/GeoData.jl/pull/168 as it requires iteration over length one tuples/ref etc.

I'm working on whole-column iteration to allow broadcasting over a longer tuple, fixing #37 and a lot of other things. But its a lot more work and I imagine may require some working through due to the increase in memory use in its simplest form.

meggart:
Yes, and sorry for getting diverted from this by other tasks. I will merge for now and we can find a more general solution later

PR #45, opened 2021-09-01
add method for Array, copyto, reverse where the fallbacks are broken

rafaqz:
This PR adds a few more base array methods - `Array`, `copyto!` and `reverse` that are currently broken with chunks and the fallback implementation. `Array` and `copyto` use broadcast `=`, while `reverse` uses `view`.

Also `replace` uses broadcast instead of a loop, which means that `count` doesn't work, but that is not a common use case anyway.


rafaqz:
Any comments on this @meggart?

PR #46, opened 2021-12-22
fix reshape with existing trailing 1s

rafaqz:
Fixes reshape breaking when there are existing trailing 1s in the array.

See: https://github.com/rafaqz/Rasters.jl/issues/234


meggart (approved):
Thanks a lot for the fix

PR #47, opened 2022-01-07
Redesign of GridChunks type

meggart:
This PR implements a re-design of the GridChunks type which defines the chunk sizes for a DiskArray. The main purpose of this redesign is to provide more first-class support for arrays with non-constant chunk sizes. This happens quite frequently when concatenating NetCDFs along an axis that does not have constant length (for example concatenating annual NetCDF files along time axis where some have a leap year and some don't). An example use case is `ConcatDiskArray` from DiskArrayTools.jl. 

This PR makes it possible to use views, reshape, broadcast and permutations also on these irregular arrays. Since we change the way that `GridChunks` works, I would bump the minor version here although that name is not exported. This would also give us a bit more time for consolidation and testing and dependents can decide when they want to update their deps sections and check if tests pass. @rafaqz it would be nice if you can test this with a recent DimensionalData and Rasters version as well

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.3' # Replace this with the minimum Julia version that your package supports. E.g. if your package requires Julia 1.5 or higher, change this to '1.5'.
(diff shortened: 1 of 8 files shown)


meggart:
I have tested this update with several downstream packages and so far everything seems ok, so I would like to go ahead and merge, if there are no objections.

rafaqz:
Sorry I've been on the road a bit. I've tested with Rasters.jl and had no problems.

PR #55, opened 2022-04-12
apply bluestyle

rafaqz:
This package is by necessity somewhat complicated and difficult to understand, but also an important dependency it would be good if more people could contribute to.

To ease the pain of learning it a little a little this PR applies a consistent style (https://github.com/invenia/BlueStyle) that is a bit more familiar to most julia developers.

Essentially this means:
- choose four space indent, currently we have 2 and 4
- spaces after commas and hashes
- spaces around infix operators other than colon
- explicit `return` in multi line functions
- a few other minor details

I also did some organising into blocks of Base/DiskArrays methods. 

A follow up PR will be to improve the documentation a little.

### DIFF
--- a/src/DiskArrays.jl
+++ b/src/DiskArrays.jl
@@ -1,232 +1,8 @@
 module DiskArrays
+
 export AbstractDiskArray, interpret_indices_disk
 
-"""
-Abstract DiskArray type that can be inherited by Array-like data structures that
-have a significant random access overhead and whose access pattern follows
-n-dimensional (hyper)-rectangles.
-"""
-abstract type AbstractDiskArray{T,N} <: AbstractArray{T,N} end
-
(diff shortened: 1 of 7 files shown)


meggart:
Great, thanks a lot. I think it would also be great to setup a JuliaFormatter Github Action, so that this cleanup can happen in an automatted way.

rafaqz:
For sure. I didn't actually use JuliaFormatter this time I guess that would be better. I would also like to swap all internal method names to start with underscores to make that consistent too.

rafaqz:
For sure. I didn't actually use JuliaFormatter this time I guess that would be better.

A lot of this was layout and rearranging functions into blocks and separated files which auto formatting wont do. There are a few more things like that left, like using an underscore prefix on internal methods.

PR #56, opened 2022-04-13
allow mixed sized chunks

### LINKED ISSUE #50: With changes to chunking, `copyto!` now fails in places it used to succeed.
Writing with GDAL to a non-chunked file like ".rst" from a chunked file like ".tif" gives the dreaded 

> `Chunks do not align in dimension 2` 

error.

This is an unfortunate effect of the recent changes. I'm not sure of the exact reason.

rafaqz:
This PR defines a method that finds the smallest compatable chunk size in each dimension, in the case that a "Chunks do not allign" error would have been thrown. It wont handle all use cases, but deals with the case that one chunk size is a multiple of the other.

Seems to work fine, I just need to write some tests but @meggart I'm creating a PR now so you don't bother looking at it in the meantime. 

Closes #50


rafaqz:
@meggart before I work on this further - is your `batch_getindex` branch achieving a similar thing?

meggart:
And thanks a lot for starting this. I have actually worked on this issue in a different context, but rather trying to find a fast solution that works on large data, where I used Optim.jl to find an ideal processing chunk size for broadcasting. However, this is in very ealry stages and needs more expemerintation, so I guess I might have something in a year or so. 

And even then, I don't think this should be part of DiskArrays because of the quite heavy dependency. However, if there was a way to make the actual broadcast execution generic and extensible, this would be great. Then we might also think about multi-threaded or distributed execution of a broadcast expression, or one handled by Dagger etc. 
So some way for packages to determine how `Base.copyto!(dest::AbstractArray, bc::Broadcasted{ChunkStyle{N}})` works would be great. I am not asking you to implement this just wondering if this might be a good idea.

rafaqz:
Thanks for the explanation, the batch indexing would also be a good thing to have. It might just clash with this code a little!

This is a very simple solution. I just loop over the block ranges for each dimension bumping each until one of the block sizes for all the arrays match. 

It should work for any combination of dimensions, but it definitely doesn't find an optimum size! just a size that wont error. Which I think is the more common use case anyway. Currently if there are no common block boundaries it will just load the whole array and fail via memory use for a large file. Probably we need to cap the maximum size to e.g. half the remaining RAM before that happens.

rafaqz:
@meggart this should be good to merge. It solves the problem of where chunks are different sizes but multiples of each other, removing the most common current chunk error. That's enough to bump compat in Rasters.jl

It leaves the harder problem of totally misaligned chunks (such as from views over different chunk sizes). This requires choosing which blocks are the least cost to load twice, and a slightly more complicated algorithm to track which block ranges to look in.

src/DiskArrays.jl
```diff
@@ -6,6 +6,7 @@ include("diskarray.jl")
 include("array.jl")
 include("chunks.jl")
 include("broadcast.jl")
+include("mapreduce.jl")
 include("iterator.jl")
 include("mapreduce.jl")
```
meggart, line 11: I think mapreduce.jl is included twice here

meggart (approved):
Thanks a lot for the changes, as I said I am doing a lot of experimental work on matching non-aligned chunks. However, until this is done your solution at at least works for small files + for large files when the chunk sizes match approximately.

PR #62, opened 2022-05-09
Subsetchunks

### LINKED ISSUE #61: `view` with an AbstractRange (not AbstractUnitRange) is broken
This is because `subsetchunks` is specifically only written for `reverse`

https://github.com/meggart/DiskArrays.jl/blob/d0da1e0f4bae3f464b873f1659689c6ece398a18/src/chunks.jl#L43-L50

Should we just subset it as if there is not step? will that work?

Edit: it doesn't.

Maybe the batch PR can work for this as well? views like this are very useful for reducing the size of (very) large rasters for plotting. Rasters.jl has been using this prior to 0.3.

meggart:
Add methods to subsetchunks to fix #61


meggart:
Please don't merge yet. I will still add more tests to make sure we hit all branches

meggart:
Ok, I think this is thoroughly tested now and views with steps >1 work again. However, working with these kinds of views might still be inefficient until I finish #59 . This depends on if the DiskArray in question has a custom implementation for `readblock!(a, aout, i::OrdinalRange...)` which is optimised to read strided data from the underlying source. Otherwise a fallback will be hit which reads the bounding box of the range and does the subsetting afterwards, which unexpectedly might blow up IO and memory usage. I will make a few test cases in #59 to make sure this is improved.

PR #65, opened 2022-06-07
add tests for permutedims and batch indices

gdkrmr:
I just wanted to add some quick tests for the permutedims case and this breaks other tests. In don't think any of the methods should change the data in place.


rafaqz:
I tend to agree. You are welcome to fix that as well.

gdkrmr:
I have "fixed" the test and set the number to 10. I think this is the correct number because we have to load 4 chunks.

meggart (approved):

meggart:
Thanks a lot @gdkrmr

PR #69, opened 2022-08-11
Set up documentation

rafaqz:
Needed some tweaks to the readme badges so the readme could also work as module docs.

We still also need to set up documenter keys for github actions, I don't have access to do this in settings.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -1,0 +1,26 @@
+name: Documentation
+
+on:
+  push:
+    branches:
+      - main
+    tags: '*'
+  pull_request:
+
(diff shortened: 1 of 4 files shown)


rafaqz:
Seems like the formatting bot has broken master somehow... that shouldn't be possible??

rafaqz:
@meggart would it be possible to add a DOCUMENTER_KEY to this repo so we can get the documentation working? I don't have access to that.

meggart:
I think this should work using the GITHUB_TOKEN as well. I think the reason here is that branch is still called `master` while the Documenter.yml file is listening to changes to `main`. I will rename the branch

PR #74, opened 2022-08-28
add an `allow_scalar` method

rafaqz:
Introduce an optional check for scalar indexing as in CUDA.jl, using the CUDA.jl syntax.

This should help identify causes of performance bugs in future, e.g. in issues like https://github.com/rafaqz/Rasters.jl/issues/298

### DIFF
--- a/src/DiskArrays.jl
+++ b/src/DiskArrays.jl
@@ -9,6 +9,7 @@ end DiskArrays
 
 export AbstractDiskArray, interpret_indices_disk, eachchunk, ChunkIndex, ChunkIndices
 
+include("scalar.jl")
 include("diskarray.jl")
 include("array.jl")
 include("broadcast.jl")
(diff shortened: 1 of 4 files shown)


meggart (approved):
This is a great idea and I agree this will help finding bugs in the future.

meggart:
Do you have an idea why the tests were not executed by GH actions?

meggart:
Unrelated, but I just looked at the issue you linked, there is already an implementation of `ConcatDiskArray` here (for concatenating along existing dimensions): https://github.com/meggart/DiskArrayTools.jl/blob/16539572a5042f272d4d74adbbcec928f1862ceb/src/DiskArrayTools.jl#L203
and one for `StackedDiskArray` for stacking along new dimensions: https://github.com/meggart/DiskArrayTools.jl/blob/16539572a5042f272d4d74adbbcec928f1862ceb/src/DiskArrayTools.jl#L12 which might be useful to implement a generic `cat`

rafaqz:
Ok thatts great, also useful for some other issues like https://github.com/rafaqz/Rasters.jl/issues/242

Should we move those here so we can define `cat`?

rafaqz:
It seems we're accessing a non-API field in iterators.jl, on nightly there is the error  `type Stateful has no field taken`

https://github.com/meggart/DiskArrays.jl/runs/8178960676?check_suite_focus=true#step:6:101

PR #81, opened 2022-11-26
use length for chunk iterator state rather than .taken

rafaqz:
Fixes #80 and errors on nightly

PR #82, opened 2022-12-19
bump patch version

visr:
Would be great to have #81 released such that ArchGDAL works on 1.9.


rafaqz:
Oops I mistook this for an issue and made another PR :sweat_drops:

PR #98, opened 2023-05-19
Fix 97

meggart:


src/diskarray.jl
```diff
@@ -22,9 +22,22 @@ should be supported as well.
 """
 function writeblock!() end
 
+# This is for filtering "true" batch getindex functions with vector indexing from
+# excess singleton dimensions with values like [1] and 1:1
+function is_batch_arg(j::AbstractArray)
+    if length(j) == 1
+        first(j) == 1 ? false : throw(ArgumentError("Singleton dimension value is not equal to 1"))
```
rafaqz, line 29: What about selecting a range like `2:2`? shouldn't that work?

Do we need to instead check that _trailing_ indices are equal 1? But maybe I don't totally understand whats happening here.

meggart: Thanks a lot, good catch. I was mixing the trailing argument check and the batchgetindex here

PR #99, opened 2023-07-07
Fix Zerodimensional broadcast bug

meggart:
Fixes https://github.com/rafaqz/Rasters.jl/issues/473


meggart:
Sorry I messed up my branches, the first commit was not supposed to be in here

PR #102, opened 2023-07-23
add lazy concatenating disk arrays and Base.cat method

rafaqz:
This is half code modified from DiskArrayTools.jl

It adds a `ConcatDiskArray` wrapper, and a `cat` method that applies it.

We could possibly try to merge if a `ConcatDiskArray` is passed to the `ConcatDiskArray` constructor?

### DIFF
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -138,8 +138,12 @@ macro implement_cat(t)
     t = esc(t)
     quote
         # Allow mixed lazy cat of other arrays and disk arrays to still be lazy
+        # TODO this could be better. allowing non-AbstractDiskArray in
+        # the macro makes this kind of impossible to avoid dispatch problems
         Base.cat(A1::$t, As::AbstractArray...; dims::Int) = cat_disk(A1, As...; dims)
         Base.cat(A1::AbstractArray, A2::$t, As::AbstractArray...; dims::Int) = cat_disk(A1, A2, As...; dims)
         Base.cat(A1::$t, A2::$t, As::AbstractArray...; dims::Int) = cat_disk(A1, A2, As...; dims)
+        Base.vcat(A1::Union{$t{<:Any,1},$t{<:Any,2}}, As::Union{$t{<:Any,1},$t{<:Any,2}}...) = cat_disk(A1, As...; dims=1)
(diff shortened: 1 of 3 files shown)


src/cat.jl
```diff
@@ -0,0 +1,135 @@
...
+            max(first(indstoread) - si[ii] + 1, 1):min(last(indstoread) - si[ii] + 1, ms)
+        end
+        outer_range = map(cI.I, a.startinds, array_range, inds) do ii, si, ar, indstoread
+            (first(ar)+si[ii]-first(indstoread)):(last(ar)+si[ii]-first(indstoread))
+        end
+        aout[outer_range...] = a.parents[cI][array_range...]
+    end
+end
+
+function writeblock!(a::ConcatDiskArray, aout, inds::AbstractUnitRange...)
```
meggart, line 85: Do you think we need this? It would be more or less copy/paste of `readblock!` but with LHS and RHS reversed in line 81. So in principle better to make a generic IO function but we can defer this to later as well. 

The main use case would be if someone wanted to write results into a pre-generated annual array of netcdf files or similar.

meggart (approved):
Thanks a lot for porting this over. Would be great to have a generic `cat` working.

meggart:
> We could possibly try to merge if a ConcatDiskArray is passed to the ConcatDiskArray constructor?

This would be a nice optimization. So far in my YAXArrays code I was usually using the direct constructor passing an array of arrays. However, since it is a common pattern to use things like `reduce(cat,array_of_diskarrays)` we should make sure to not have 1000 nested `ConcatDiskArrays`.

src/cat.jl
rafaqz: That sounds useful. I'll move the `readblock!` code to a function and call it in botg

rafaqz:
I might leave the unwrapping `ConcatDiskArray` part to another time, but I implemented `writeblock!` and tested mixed chunk sizes

meggart (approved):

meggart:
Ok, thanks a lot. The nested ConcatDIskArrays problem I can take care of as well. So feel free to merge and tag a new version.

PR #105, opened 2023-08-05
fix zip

### LINKED ISSUE #103: Fix zip and other cases of simultaneous iteration
`zip` is used in array equality, and breaks chunking order when arrays with different chunks are used.

We should at minimum fix it so that the chunks of the first disk array found are used for all others (and for regular arrays). This can be achieved with a `RechunkDiskArray` wrapper that simply forces a different chunk pattern. It may be slow in some cases but will at least be correct.

Later we could optimise the combined iteration pattern of all diskarrays passed to `zip`.

rafaqz:
This PR fixes `zip` by adding a `RechunkedDiskArray` that we can use to force a specific access pattern on a `AbstracDiskArray` or a regular `AbstractArray`.

In `zip` we then force all the arrays to have the same pattern as the first disk array we find that is `Chunked`.

This will be a bit slower when the chunks don't match, but at least it is correct. We could later calculate some optimal chunk pattern for all arrays passed to `zip` given the available memory.

This fixes some outstanding `@test_broken` tests, because `==` uses zip underneath.

Closes #103

### DIFF
--- a/src/rechunk.jl
+++ b/src/rechunk.jl
@@ -1,0 +1,16 @@
+# Force disk any abstractarray into a different chunking pattern.
+# This is useful in `zip` and other operations that can iterate
+# over multiple arrays with different patterns.
+
+struct RechunkedDiskArray{T,N,A<:AbstractArray{T,N},C} <: AbstractDiskArray{T,N}
+    parent::A
+    chunks::C
+end
+
+Base.parent(A::RechunkedDiskArray) = A.parent
+Base.size(A::RechunkedDiskArray) = size(parent(A))
+Base.getindex(A::RechunkedDiskArray, I...) = getindex(parent(A), I...)
+Base.setindex!(A::RechunkedDiskArray, I...) = setindex!(parent(A), I...)
+
+haschunks(::RechunkedDiskArray) = Chunked()
+eachchunk(A::RechunkedDiskArray) = A.chunks
--- a/src/rechunk.jl
+++ b/src/rechunk.jl
@@ -9,8 +9,19 @@ end
 
 Base.parent(A::RechunkedDiskArray) = A.parent
 Base.size(A::RechunkedDiskArray) = size(parent(A))
-Base.getindex(A::RechunkedDiskArray, I...) = getindex(parent(A), I...)
-Base.setindex!(A::RechunkedDiskArray, I...) = setindex!(parent(A), I...)
+# These could be more efficient with memory in some cases, but this is simple
+readblock!(A::RechunkedDiskArray, data, I...) = _readblock_rechunked(A, data, I...)
+readblock!(A::RechunkedDiskArray, data, I::AbstractVector...) = _readblock_rechunked(A, data, I...)
+writeblock!(A::RechunkedDiskArray, data, I...) = writeblock!(parent(A), data, I...)
 
 haschunks(::RechunkedDiskArray) = Chunked()
 eachchunk(A::RechunkedDiskArray) = A.chunks
+
+function _readblock_rechunked(A, data, I...)
+    if haschunks(parent(A)) isa Chunked
+        readblock!(parent(A), data, I...)
+    else
+        # Handle non disk arrays that may be chunked for e.g. chunked `zip`
+        copyto!(data, view(parent(A), I...))
+    end
+end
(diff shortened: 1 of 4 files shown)


rafaqz:
I updated this to `collect` in the right order using a `DiskZip` wrapper.

Currently it will only work with other `AbstractArray` iterators the same size as the disk array.

src/zip.jl
```diff
@@ -0,0 +1,79 @@
...
+    end
+end
+# For now we only allow zip on exact same-sized arrays
+DiskZip(As...) = throw(ArgumentError("zip on disk arrays only works with other same-sized AbstractArray"))
+
+# Collect zipped disk arrays in the right order
+function Base.collect(dz::DiskZip)
+    out = similar(first(dz.is), eltype(dz))
+    i = iterate(dz)
+    for I in eachindex(first(dz.is))
```
meggart, line 48: Would it be possbile to do a block-wise copying here instead? I think we would need some nested iterator here, similar to what I defined in Iterators.jl. Maybe we can rather fix that one and adapt, because this implementation here will have plenty of random accesses and will be horribly slow.

test/runtests.jl
```diff
@@ -346,6 +346,29 @@ end
...
+    a = collect(da)
+    zd = zip(da, da)
+    za = zip(a, a)
+    @test collect(zd) == collect(za)
+    @test all(zd .== za)
+    a = collect(da)
+    @test zip(a, da, a) isa DiskArrays.DiskZip
+    @test zip(da, da, a) isa DiskArrays.DiskZip
+    @test zip(da, da, da) isa DiskArrays.DiskZip
+    @test zip(a, da, da) isa DiskArrays.DiskZip
```
meggart, line 360: It would be really good to check how many getindex/setindex calls are done into the underlying arrays. Could we try to diagnose the `getindex_count(da)` to understand what is happening?

src/rechunk.jl
```diff
@@ -0,0 +1,16 @@
+# Force disk any abstractarray into a different chunking pattern.
+# This is useful in `zip` and other operations that can iterate
+# over multiple arrays with different patterns.
+
+struct RechunkedDiskArray{T,N,A<:AbstractArray{T,N},C} <: AbstractDiskArray{T,N}
```
meggart, line 5: I really support adding this struct, it would have been useful in many occasions, and we might also re-formulate the basic broadcast using this.

meggart (changes requested):
Thanks a lot for starting to tackle these correctness issues. I am still a bit concerned about the potetnial performance here, I will do some tests by myself to really understand why zip is currently broken and maybe I come up with a fix that requires less code. I will comment soon.

src/rechunk.jl
```diff
@@ -0,0 +1,16 @@
...
+    parent::A
+    chunks::C
+end
+
+Base.parent(A::RechunkedDiskArray) = A.parent
+Base.size(A::RechunkedDiskArray) = size(parent(A))
+Base.getindex(A::RechunkedDiskArray, I...) = getindex(parent(A), I...)
```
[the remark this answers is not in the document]
rafaqz: No reason  I'll change it

meggart:
> Let me know when you are finished with your changes, I can look regarding `readblock!` and `writeblock!`, I will add some performance checks then and test the getindex_counts etc.

src/rechunk.jl
rafaqz: I had to look around to make sure we didn't have it already, it seems like such a basic component. But I hadn't thought about it either before zip.

test/runtests.jl
rafaqz: Will have a look

src/zip.jl
rafaqz: Turns out its fine as is, the chunking just propagates through.

test/runtests.jl
rafaqz: I added `getindex_count` checks to the tests, its only `6` per array - so only chunked reads

src/zip.jl
rafaqz: Its amazing actually it seems all the iterators just composes properly though `DiskZip`, `Zip` and each individual stateful iterator. Without doing anything special here.

rafaqz:
This may be fine to merge now as chunking was actually working already, but if there may be some other optimisations that could be done if you want to make any changes

meggart (approved):

meggart:
Yes, I had not realized this was already working. Thanks a lot for adding the test. Is this merge conflict critical?

PR #114, opened 2023-09-17
fix iteration performance

rafaqz:
A few fixes for iteration performance, mostly gridchunks having an abstract type field was causing a lot of overhead.

There is still a pretty big cost from Iterators.Stateful for  `Unchunked()` arrays that don't actually need it. We can probably fix that as well at some stage.

(I moved the code into `_iterate_disk` so we can do more of this work in future more easily)


meggart (approved):
Looks good to me

PR #124, opened 2023-10-05
make 1D variable with an unlimited dimensions behave in the same way as 2D variables

Alexander-Barth:
This solves issue https://github.com/meggart/DiskArrays.jl/issues/123 for NetCDF.jl and NCDatasets.jl

### DIFF
--- a/src/diskarray.jl
+++ b/src/diskarray.jl
@@ -116,6 +116,13 @@ function interpret_indices_disk(
     end
 end
 
+
+function interpret_indices_disk(A::AbstractVector, r::NTuple{1,AbstractVector})
+    inds = map(_convert_index, r, size(A))
+    resh = DimsDropper(findints(r))
+    return inds, resh
+end
+
 # function interpret_indices_disk(A, r::Tuple{<:AbstractArray{<:Bool}})
 #   ba = r[1]
 #   if ndims(A)==ndims(ba)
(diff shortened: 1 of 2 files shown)


rafaqz:
Ok that makes sense, it was trying to handle linear indices on a vector.

Can you write a test so we dont break this again?

Alexander-Barth:
Good idea! Would it be ok to add NetCDF.jl as a test dependency? I am not sure how to create such array using only DiskArrays.

rafaqz:
There is a `_DiskArray` object defined in the tests that we use. It counts `getindex` so you cant test the read pattern (e.g. you want one single read here), see the tests for examples.

Alexander-Barth:
OK, I just added the test. Without commit 76152b8cbd96fa07c4761a0dc1e442b7a9df3089 I could reproduce the failure in the function `interpret_indices_disk` which is fixed by this PR.

rafaqz:
Looks good! Thanks for adding that.

Would it make sense to explicitly test 2d arrays there too?

meggart:
Thanks a lot. As you @rafaqz mentioned there was a special method for linear indexing into multidimensional arrays and in this case it was taking over linear indexing into 1D arrays as well. Looks good to me.

Alexander-Barth:
OK, I just added a 2d array.

PR #128, opened 2023-10-08
allow all Integer types in RegularChunks

rafaqz:
ArchGDAL is broken on master because it passes Int32 to GridChunks


rafaqz:
@meggart I'm going to merge this asap so we can bump the patch version and avoid people hitting the bug in ArchGDAL

Issue #131, opened 2023-10-20
optimization for var[:,:,[1,2]], wrong result with indexing with two vectors and the type-stability issue


Alexander-Barth:
In NCDatasets 0.12.17 I had optimization for `var[:,:,[1,2]]` converting it to `var[:,:,1:2]` and
`var[:,:,[1,2,10,11]]` into two load operation `v[:,:,1:2]` and `v[:,:,10:11]`. This was implemented here:

https://github.com/Alexander-Barth/NCDatasets.jl/blob/v0.12.17/src/cfvariable.jl#L359-L436

Now a call to `v[:,:,[1,2]]` loads every element individually (which quite slow).

In fact, the relevant code is still in NDatasets 0.13 but no longer called.

Maybe this functionality should go into DiskArrays?

rafaqz:
PRs always welcome

Alexander-Barth:
OK, DiskArrays has implementation of "batch" getindex.jl
https://github.com/meggart/DiskArrays.jl/blob/7d297f8b4a0067f7c93cb84e8c2d42b88754e54e/src/batchgetindex.jl

Because of too restrictive type declaration in CommonDataSet, I was not using it. They are now fixed. Now I am using batchgetindex.jl but there are two issues in its implementation:

With the `_DiskArray` type from runtests.jl I made these two tests:

```julia
a1 = _DiskArray(zeros(2, 6000));
@show getindex_count(a1)
a1[:,[1,6000]]
@show getindex_count(a1)
```

output

```
getindex_count(a1) = 0
┌ Info: readblock! of _DiskArray
└   i = (1:2, 1:6000)
getindex_count(a1) = 1
```

The `i` variable in the parameter to readblock!.
Surprisingly, here I get only 1 call the getindex but loading quite a lot of data for this case (in fact all). I wondering if this is intentional?

The other issue is that the output can be wrong in some circumstances:

```julia
a1 = _DiskArray(zeros(5,5,10));
size(a1[[1,2],[2,3],:])
# output (2, 3, 2)
# expected (2,2,3)
```

The code in batchindex is really complicated. I am wondering if we should replace it with my implementation (loads less data and works if there are two indices as vectors)

Alexander-Barth:
Another issue I just came accross is that with DiskArray the indexing is no-longer type stable (in this case):

Currently NCDatasets + DiskArrays:

```
julia> v = NCDataset("/tmp/sample2.nc")["data"].var;

[...]

PR #132, opened 2023-10-24
[WIP] issue DiskArrays.jl #131

Alexander-Barth:
In this PR, fixes the following issue 

```julia
a1 = _DiskArray(zeros(5,5,10));
size(a1[[1,2],[2,3],:])
# previous output (2, 3, 2)
# now  (2,2,3)
```

also it is now type stable

```
ulia> using NCDatasets; v = NCDataset("/tmp/sample2.nc")["data"].var;
[ Info: Precompiling NCDatasets [85f8d34a-cbdd-5861-8df4-14fed0d494ab]

julia> foo(v) = v[:,:,[1]];

julia> @code_warntype foo(v)
MethodInstance for foo(::NCDatasets.Variable{Float32, 3, NCDataset{Nothing}})
  from foo(v) @ Main REPL[2]:1
Arguments
  #self#::Core.Const(foo)
  v::NCDatasets.Variable{Float32, 3, NCDataset{Nothing}}
Body::Array{Float32, 3}
1 ─ %1 = Main.:(:)::Core.Const(Colon())
│   %2 = Main.:(:)::Core.Const(Colon())
│   %3 = Base.vect(1)::Vector{Int64}
│   %4 = Base.getindex(v, %1, %2, %3)::Array{Float32, 3}
└──      return %4

julia> @which v[:,:,[1]]
getindex(a::NCDatasets.Variable, i...)
     @ NCDatasets ~/.julia/dev/DiskArrays/src/diskarray.jl:211

```

It avoids also reading too much of data.

I had to set the `need_batch = false` to avoid infinite recursion. Any advice would be appreciated.

The implementation in NCDatasets is added to current implementation of the special case:

```
batchgetindex(a::TA,indices::Vararg{Union{Int,Colon,AbstractRange{<:Integer},Vector{Int}},N}) where TA <: AbstractArray{T,N} where {T,N}
```

to that we do not remove any functionality.

### DIFF
--- a/src/batchgetindex.jl
+++ b/src/batchgetindex.jl
@@ -67,6 +67,91 @@ function batchgetindex(a, i::AbstractVector{Int})
     ci = CartesianIndices(size(a))
     return batchgetindex(a, ci[i])
 end
+
+# indexing with vector of integers from NCDatasets 0.12.17 (MIT)
+
+# computes the shape of the array of size `sz` after applying the indexes
+# size(a[indexes...]) == _shape_after_slice(size(a),indexes...)
+
+# the difficulty here is to make the size inferrable by the compiler
+@inline _shape_after_slice(sz,indexes...) = __sh(sz,(),1,indexes...)
+@inline __sh(sz,sh,n,i::Integer,indexes...) = __sh(sz,sh,               n+1,indexes...)
+@inline __sh(sz,sh,n,i::Colon,  indexes...) = __sh(sz,(sh...,sz[n]),    n+1,indexes...)
+@inline __sh(sz,sh,n,i,         indexes...) = __sh(sz,(sh...,length(i)),n+1,indexes...)
+@inline __sh(sz,sh,n) = sh
+
+
+# convert e.g. vector indices to a list of ranges
+# [1,2,3,6,7] into [1:3, 6:7]
+# 1:10 into [1:10]
+to_range_list(index::Integer,len) = index
+to_range_list(index::Colon,len) = [1:len]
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -493,7 +492,8 @@ end
     #Index with range stride much larger than chunk size
     a = _DiskArray(reshape(1:100, 20, 5, 1); chunksize=(1, 5, 1))
     @test a[1:9:20, :, 1] == trueparent(a)[1:9:20, :, 1]
-    @test getindex_count(a) == 3
+    # now getindex_count(a) == 1
```
rafaqz, line 495: Doesn't this mean sparse ranges will now read all the data in some cases?

I thought the ideas of `batchgetindex` was to exlicitly just read the required chunks.

src/batchgetindex.jl
```diff
@@ -171,6 +256,8 @@ function _readblock!(A::AbstractArray, A_ret, r::AbstractVector...)
         mi, ma = extrema(ids)
         return largest_jump > cs && length(ids) / (ma - mi) < 0.5
     end
+    # What TODO?: necessary to avoid infinite recursion
+    need_batch = false
```
rafaqz, line 260: This is essentially deleting half the code in this file, there may be a case for reorganising things but I think a lot of optimisations are being thrown out doing this.

meggart:
I think the issues mentioned here have been fixed in other PRs

Alexander-Barth:
Thanks!

PR #134, opened 2024-01-10
Dispatch on the dims type in cat_disk to allow for one element tuples for dims in cat

### LINKED ISSUE #133: `cat` should work for iterators of Ints
According to the documentation of `cat` it should be possible to call it with an iterable of dimensions so that the "the positions along these dimensions are increased
  simultaneously for each input array, filling with zero elsewhere. This allows one to construct
  block-diagonal matrices as cat(matrices...; dims=(1,2)), and their higher-dimensional analogues."

We should try to adhere to this interface and I am wondering whether we would need to have a new type ConcatBlockDiskArray to overload the writeblock and readblock functions.

This part of the cat interface is used for example in DimensionalData. Therefore we can't currently use cat of a DiskArray backed AbstractDimArray.

felixcremer:
This would partially fix #133 by handling the case when the iterator has only one element and that needs to put out of the tuple and forward to the original method. This makes it possible to use cat from DimensionalData with DiskArray backed arrays.

I had to rearrange the cat_disk function so that we can dispatch on the dims type. 

The entries in the Project.toml are because Aqua was complaining.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,10 @@ version = "0.3.22"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 
 [compat]
+Aqua = "0.8"
 OffsetArrays = "1"
+Statistics = "1.6"
+Test = "1.6"
 julia = "1.6"
 
 [extras]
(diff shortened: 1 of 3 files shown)


rafaqz (approved):
LGTM, kind of embarrassing it doesnt work with DD 😂

PR #143, opened 2024-01-31
use summary and eachchunk in show

rafaqz:
This PR adds a `@implement_show` macro so show can be applied to non-`AbstractDiskArray`, and shows the type `summary` and chunk type from `haschunks`.

For `Chunked` it also shows the chunk pattern - I'm not sure we need this bit but its helpful sometimes.

It looks like this:
```julia
julia> a = AccessCountDiskArray(data; chunksize=(5, 4, 2))
10×20×2 AccessCountDiskArray{Float64, 3, Array{Float64, 3}}

Chunked: 2×5×1 DiskArrays.GridChunks{3, Tuple{DiskArrays.RegularChunks, DiskArrays.RegularChunks, DiskArrays.RegularChunks}}:
[:, :, 1] =
 (1:5, 1:4, 1:2)   (1:5, 5:8, 1:2)   (1:5, 9:12, 1:2)   (1:5, 13:16, 1:2)   (1:5, 17:20, 1:2)
 (6:10, 1:4, 1:2)  (6:10, 5:8, 1:2)  (6:10, 9:12, 1:2)  (6:10, 13:16, 1:2)  (6:10, 17:20, 1:2)
```

### DIFF
--- a/src/DiskArrays.jl
+++ b/src/DiskArrays.jl
@@ -22,6 +22,7 @@ include("rechunk.jl")
 include("cat.jl")
 include("generator.jl")
 include("zip.jl")
+include("show.jl")
 
 # The all-in-one macro
 
@@ -41,6 +42,7 @@ macro implement_diskarray(t)
         @implement_batchgetindex $t
         @implement_cat $t
         @implement_zip $t
+        @implement_show $t
         @implement_generator $t
     end
 end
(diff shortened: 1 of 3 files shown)


rafaqz:
This will need a minor version bump because show could clash with existing methods and thats an error now after 1.10

PR #151, opened 2024-02-20
Allow indexing with nd arrays of integers and CartesianIndex

### LINKED ISSUE #84: dimensions get moved around in unintuitive ways when accessing DiskArray using CartesianIndex
I noticed the the dimensions get moved around in unintuitive ways

````julia
dc = Zarr.zopen(path2zarr)

size(dc["v"])
(834, 834, 84396)

r = [1 4 6 1]
c = [7 8 9 3]

cartind = CartesianIndex.(r,c)

size(dc["v"][cartind, :])
(1, 84396, 4)

cartind = CartesianIndex.(r',c')

size(dc["v"][cartind, :])
(4, 84396, 1)
````

This is different behavior that non DiskArrays
````julia
z = rand(100,100,1000)

cartind = CartesianIndex.(r,c)

size(z[cartind,:])
(1, 4, 1000)

cartind = CartesianIndex.(r',c')
size(z[cartind,:])
(4, 1, 1000)
````

### PR DESCRIPTION
**Pull request by meggart**
This fixes cases where some indices are matrices of integers or CartesianIndex so that the output has the correct shape afterwards. Fixes #84

### DIFF
--- a/src/batchgetindex.jl
+++ b/src/batchgetindex.jl
@@ -67,8 +67,8 @@ end
 has_chunk_gap(cs,ids) = true
 
 #Compute the number of possible indices in the hyperrectangle
-span(v::AbstractVector{<:Integer}) = 1 -(-(extrema(v)...))
-function span(v::AbstractVector{CartesianIndex{N}}) where N
+span(v::AbstractArray{<:Integer}) = 1 -(-(extrema(v)...))
+function span(v::AbstractArray{CartesianIndex{N}}) where N
     minind,maxind = extrema(v)
     prod((maxind-minind+oneunit(minind)).I)
 end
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/batchgetindex.jl [hunk 1]
```diff
@@ -110,7 +111,7 @@ function process_index(i::AbstractVector{<:Integer}, cs, ::ChunkRead)
         push!(tempinds, (tempind,))
         maxtempind = max(maxtempind,maximum(tempind))
     end
-    (length(i),), ((maxtempind),), (outinds,), (tempinds,), (datainds,), Base.tail(cs)
+    size(i), ((maxtempind),), (outinds,), (tempinds,), (datainds,), Base.tail(cs)
```
rafaqz, line 114: This  return tuple is so long, should we make it a `NamedTuple` or even make a struct for it?

meggart: Yes, in principle it might be better to return a struct here. However, I am quite happy that currently everything infers well and am a bit reluctant to touch this part right now, since these process_index chains are the most demanding parts for the compiler. Could we fix this later?

meggart: I am preparing another PR that introduces some DiskIndex type, which I will also expand to include some information on aliasing between temp and output array. So I will merge this as is for now

PR #161, opened 2024-03-05
Avoid allocation of temp array for dense data reads and writes

meggart:
This is a mere optimization and tries to avoid allocating a temporary array wherever possible during getindex/setindex calls.

### DIFF
--- a/src/diskarray.jl
+++ b/src/diskarray.jl
@@ -41,8 +41,8 @@ Determines a list of tuples used to perform the read or write operations. The re
 - `data_indices` indices for reading from data array
 """
 Base.@assume_effects :removable resolve_indices(a,i) = resolve_indices(a,i,batchstrategy(a))
-Base.@assume_effects :removable resolve_indices(a, i, batch_strategy) = _resolve_indices(eachchunk(a).chunks, i, DiskIndex((),(),(),(),(),PerfectAlign()), batch_strategy)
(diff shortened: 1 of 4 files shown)


meggart:
@rafaqz I don't know if you would like to have a look. This PR also includes the DiskIndex struct you were suggesting instead of returning the long tuple from process_index. In addition I extended the getindex and setindex tests so we test combinations of all types of different indexing, which revealed a few additional hidden bugs. I also completely removed the `assume_effects` macro for now, since the function infers also without this. So this passes tests on 1.9 now.

src/diskarray.jl
```diff
@@ -73,49 +71,65 @@ function need_batch_index(i, cs, batchstrat)
...
+    DiskIndex(
+        (a.output_size...,b.output_size...),
+        (a.temparray_size...,b.temparray_size...),
+        (a.output_indices...,b.output_indices...),
+        (a.temparray_indices...,b.temparray_indices...),
+        (a.data_indices...,b.data_indices...),
+    )
+end
+
+function _resolve_indices(cs, i, indices_pre::DiskIndex, nb)
```
rafaqz, line 93: Can we put a type on `nb` for all of these? would just be easier to read. Otherwise maybe spell our `nb`

rafaqz (approved):
LGTM. DiskIndex makes things a lot more organised.

PR #164, opened 2024-04-02
Fix a case of wrong aliasing for setindex

meggart:
This occured in `setindex!` cases where source and destination shape were not the same, e.g.

````julia
a[1:10,2:2] = 1:10
````

A test was added for this in setindex tests.

### DIFF
--- a/src/diskarray.jl
+++ b/src/diskarray.jl
@@ -168,7 +168,7 @@ splitcs(::Tuple{}, csnow, csrem) = (csnow, csrem)
 function output_aliasing(di::DiskIndex, ndims_dest, ndims_source)
     if all(i->isa(i,Union{Int,AbstractUnitRange,Colon}),di.temparray_indices) && 
         all(i->isa(i,Union{Int,AbstractUnitRange,Colon}),di.output_indices)
-        if di.output_size == di.temparray_size && nda == ndv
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by meggart** [comment]
For reference: This is the failing PR over at Zarr.jl https://github.com/JuliaIO/Zarr.jl/pull/140

File: src/diskarray.jl [hunk 1]
```diff
@@ -165,10 +165,10 @@ splitcs(::Tuple{}, csnow, csrem) = (csnow, csrem)
 
 #Determine wether output and temp array can a) be identical b) share memory through reshape or 
 # c) need to be allocated individually
-function output_aliasing(di::DiskIndex)
+function output_aliasing(di::DiskIndex,nda,ndv)
```
rafaqz, line 168:
```suggestion
function output_aliasing(di::DiskIndex, ndims_source, ndims_dest)
```

meggart:
And in general, I think it would be good to increase the number of unit tests for `setindex!`. A lot of work has gone into thorough testing of getindex operations and the resolution of indices goes through the same machinery for both, but some case are different, especially since the user supplies 2 arrays instead of one some rules need to be obeyed, so there are numerous combinations of additional or missing trailing indices of singleton dimensions to test, both for the array to be written to and for the array containing the data to write.

PR #169, opened 2024-04-22
Check for zero length index in copyto! and same length of the indices

### LINKED ISSUE #168: copyto! fails when one of the axes has a zero length index
Using copyto! with a zero length axis breaks with a DiskArray but succeeds with a normal Array.
This breaks the show method of DimArrays with a DiskArray in some circumstances see https://github.com/rafaqz/DimensionalData.jl/issues/670.

```julia
    a = collect(reshape(1:90, 10, 9))
    a_disk = AccessCountDiskArray(a; chunksize=(5, 3))
        x_empty = Matrix{Int64}(undef, 0,2)
        copyto!(x_empty, CartesianIndices((1:0, 1:2)), a_disk, CartesianIndices((8:7, 8:9))) # This fails
        copyto!(x_empty, CartesianIndices((1:0, 1:2)), a, CartesianIndices((8:7, 8:9))) # This works
```

felixcremer:
This checks for an index of length zero in the copyto! call so that it does not run into chunking issues further down the code. 
I am not entirely sure, whether this is the best spot to introduce this, but it seems to work. 

The check for the same length of the indices is copied from the behaviour of copyto! on normal arrays.
This will fix #168.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -69,7 +69,6 @@ function _copyto!(dest, Rdest, src, Rsrc)
 
     if any(==(0), size(Rdest))
         # This check is here to catch #168
-        println("Return dest")
         return dest
     end
     view(dest, Rdest) .= view(src, Rsrc)
--- a/src/array.jl
+++ b/src/array.jl
@@ -61,8 +61,19 @@ function _copyto!(dest::AbstractArray, source::AbstractArray)
     reshape(dest, size(source)) .= source
     return dest
 end
-_copyto!(dest, Rdest, src, Rsrc) = view(dest, Rdest) .= view(src, Rsrc)
 
+function _copyto!(dest, Rdest, src, Rsrc)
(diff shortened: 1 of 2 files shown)


src/array.jl
```diff
@@ -61,8 +61,18 @@ function _copyto!(dest::AbstractArray, source::AbstractArray)
...
     return dest
 end
-_copyto!(dest, Rdest, src, Rsrc) = view(dest, Rdest) .= view(src, Rsrc)
 
+function _copyto!(dest, Rdest, src, Rsrc)
+    if size(Rdest) != size(Rsrc)
+        throw(ArgumentError("source and destination must have same size (got $(size(Rsrc)) and $(size(Rdest)))"))
+    end
+
+    if any(==(0), size(Rdest))
```
meggart, line 70:
```suggestion
    if isempty(Rdest)
```

meggart (approved):

PR #170, opened 2024-05-12
use indexing rather than broadcast in Array

rafaqz:
This MWE is with DiskArrays.jl devved on to latest master

```julia
using BenchmarkTools
using ArchGDAL
using Rasters

fn = "deleteme.tif"
x = X(1:4096)
y = Y(1:4096)
bs = 256 # block size
rs = 512 # read size
write(fn, Raster(rand(UInt8, x,y)); options=Dict("TILED"=>"YES", "BLOCKXSIZE"=>bs, "BLOCKYSIZE"=>bs), force=true)

ext = Rasters.Extents.Extent(X=(1, rs), Y=(1, rs))
nt = (min_x=1, min_y=1, max_x=rs, max_y=rs)

r = Raster(fn; lazy=true)
v = Rasters.view(r, ext)
```

We can time indexing is 3x faster than calling `Array`. Im not really sure why.
```julia
julia> @btime $v[:, :];
  118.211 μs (130 allocations: 261.72 KiB)

julia> @btime Array($v);
  492.226 μs (518 allocations: 533.95 KiB)
```

This PR just switches `Array` over to index with all `:`. so the timings are:

```julia
julia> @btime $v[:, :];
@btime Array($v);
  118.788 μs (130 allocations: 261.72 KiB)

julia> @btime Array($v);
  123.338 μs (106 allocations: 261.42 KiB)
```

(To me this also suggest there is something wrong with our broadcasts... broadcast to a regular array should be almost identical to indexing the whole array, but instead its 3-4x slower)

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -47,9 +47,8 @@ macro implement_array_methods(t)
 end
 
 # Use broadcast to copy to a new Array
-function _Array(a::AbstractArray{T,N}) where {T,N}
-    a[ntuple(_ -> :, Val{N}())...]
-end
+_Array(a::AbstractArray{T,N}) where {T,N} = a[ntuple(_ -> :, Val{N}())...]
+_Array(a::AbstractArray{T,0}) where {T} = fill(a[])
 
 # Use broadcast to copy
 function _copyto!(dest::AbstractArray{<:Any,N}, source::AbstractArray{<:Any,N}) where {N}
--- a/src/array.jl
+++ b/src/array.jl
@@ -47,9 +47,7 @@ end
 
 # Use broadcast to copy to a new Array
 function _Array(a::AbstractArray{T,N}) where {T,N}
-    dest = Array{T,N}(undef, size(a))
-    dest .= a
-    return dest
+    a[ntuple(_ -> :, Val{N}())...]
 end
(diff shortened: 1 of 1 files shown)


rafaqz:
Hmm maybe its better to specialise `copyto!` for `Array` instead?

meggart:
The reason for the speed difference is that broadcast tries to loop over the chunks of the array doing multiple IO steps, while `a[:,:]` only performs a single IO-operation. I think specializing `copyto!` for `Array` is a very good idea, although ideally it should specialize on some `ismemory` trait. Then we could be sure that the broadcast result is "small" because the user wants to have the result in memory anyway. 

Only for broadcast operations where the target is a `DiskArray` we would need the chunk-by-chunk behavior then.

rafaqz:
Yes exactly, we should be careful about when we care about chunking. If we know we are trying to move the whole array to memory anyway we should skip it.

`ismemory` would be nice to have - it can be forwarded down through the wrappers.

rafaqz:
Can we merge this as-is for now as a stop-gap solution? and add smarter `ismemory` detection later?

meggart (approved):

PR #172, opened 2024-05-12
fix undef writes for steprange

### LINKED ISSUE #171: `setindex_disk_nobatch!` writes `undef` values for `StepRange` indices
You can see that `temparray` is defined with `undef` values, and values from `v` are written to it. But, if `i` contains `StepRanges`, temparray will still contain `undef` when it is passed to `writeblock!`:

https://github.com/meggart/DiskArrays.jl/blob/8b538f32208c8e3dbe524cf8731e2d4e766dad3e/src/diskarray.jl#L309-L311

(also note the arg order is not great in `transfer_results_write!` - we write from `v` to `temparray`.

This is the reason string writes are failing here:
https://github.com/Alexander-Barth/NCDatasets.jl/pull/247

I'm not sure what to do about it, it seems innefficient to first read the values from the dest array to fill in the block, and then write them again? but maybe it isn't?

rafaqz:
This PR fixes NCDatasets.jl for DiskArrays 0.4, and closes #171.

Currently we are writing `undef` over correct values in the array when we use a `StepRange` in `setindex!`, which is not great.

The change I made is to just pass the StepRange indices to the backend and let it handle them. Are there any cases that can hit this branch where that is not possible? (Maybe Zarr?) If so we will need a trait to mark that capability, and do the interleaving ourselves to remove the undefs for the cases where it is not.

But @meggart I'm not sure if this was anything like the intended behavior? 

Clearly it doesn't work, and needs some kind of additional conditionals/handling.

### DIFF
--- a/src/diskarray.jl
+++ b/src/diskarray.jl
@@ -261,11 +261,12 @@ function transfer_results!(o, t, oi::Tuple{Vararg{Int}}, ti::Tuple{Vararg{Int}})
     o[oi...] = t[ti...]
     o
 end
-function transfer_results_write!(v, temparray, output_indices, temparray_indices)
+
+function transfer_results_write!(temparray, v, temparray_indices, output_indices)
     temparray[temparray_indices...] = view(v, output_indices...)
     temparray
 end
-function transfer_results_write!(v, t, oi::Tuple{Vararg{Int}}, ti::Tuple{Vararg{Int}})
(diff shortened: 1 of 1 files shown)


meggart:
Thanks for starting this. I did have a look at the problem and I think the bug happens earlier in some interpret_indices method. I will investigate further...

meggart:
The issue should be fixed here #173 , so I would close this in favor of the other PR

Issue #188, opened 2024-08-27
Performance and precision of sum(ncvar_float32)


Alexander-Barth:
Prompted by [this discussion](https://discourse.julialang.org/t/ncdatasets-performance-issues-deriving-fast-minimum-maximum-value-as-too-slow/118468/2) , I had a new look at the `mapreduce` functionality (actually just the  `sum` function)  of DiskArray. 
Here are some observations that I want to share that maybe are helpful.

Here is the example NetCDF file:

```julia
fname = "/tmp/test_sample.nc"
ds = NCDataset(fname,"c")

defDim(ds,"lon",1024)
defDim(ds,"lat",1024)
defDim(ds,"time",64)
v = defVar(ds,"temperature",Float32,("lon","lat","time"),chunksizes = (512,512,1))
for n = 1:size(v,3)
    v[:,:,n] = fill(n,1024,1024)
end
close(ds)
```
With  DiskArrays v0.4.4 and NetCDF v0.12.0, I do not get the "correct" output of `sum` when applied to the NetCDF variable `temperature`:

```julia
import NetCDF

fname = "/tmp/test_sample.nc"

nc = NetCDF.open(fname);
ncvar = nc["temperature"];

@show sum(Array(ncvar))
# output 2.181038f9
# good, correct result is (64*65)/2 * 1024^2 = 2.18103808e9

res = @time sum(ncvar)
res = @time sum(ncvar)
res = @time sum(ncvar)

@show res
# output 1.0737418f9
# different from sum(Array(ncvar))

#  8.356526 seconds (469.77 M allocations: 32.250 GiB, 13.48% gc time)
```

So the output of `sum(Array(ncvar))` is  `2.181038f9` (correct) while the output of  `sum(ncvar)` is `1.0737418f9`.
Note this issue does not occur if all elements of `ncvar` are randomly initialized and ~1. I don't think that there is a bug in DiskArrays, but the rounding problem of float 32 is hitting us badly here.

Here is a very naive implementation for the sum:

```julia
using CommonDataModel
using NCDatasets
using DiskArrays


function testsum(v)
    ci = DiskArrays.eachchunk(v)

    Tdata = eltype(v)
    Tbuffer = eltype(v.var)

    N = ndims(v)
    data = Array{Tdata,N}(undef,length.(first(ci)))
    buffer = Array{Tbuffer,N}(undef,length.(first(ci)))

    res = Tdata(0)
    for indices in ci
[...]

PR #189, opened 2024-08-27
optimization for associative reductions

Alexander-Barth:
This PR addresses https://github.com/meggart/DiskArrays.jl/issues/188

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -47,3 +47,14 @@ macro implement_mapreduce(t)
         end
     end
 end
+
+
+# optimized implementation for special cases
+
+for fname in [:sum,:prod,:count,:all,:any,:minimum,:maximum]
+    @eval function Base.$fname(v::AbstractDiskArray)
+        $fname(eachchunk(v)) do chunk
+            $fname(v[chunk...])
+        end
+    end
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -907,7 +907,20 @@ end
     @test getindex_count(A) == 0 
 end
 
+@testset "optimization for associative reductions" begin
+    A = rand(1:10,30,30)
+    DA = AccessCountDiskArray(A,chunksize=(2,2))
 
+    for fun in (sum, prod, maximum, minimum)
+        @test fun(A) == fun(DA)
+    end
+
+    Ab = rand(Bool,30,30)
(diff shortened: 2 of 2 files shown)


rafaqz:
I think we also need the version with a function as first argument.

And also is count really in that list? Especially with the function? I thought it would need some tweaking.

rafaqz:
Oh right the dispatch on internals may have broken, sorry I should have realised that. We also just switched this to the main method in DimensionalData.jl

https://github.com/meggart/DiskArrays.jl/blob/ccd3092ca6f32c2663ded1d74176aa9dd222a5ef/src/mapreduce.jl#L7-L12

meggart:
> But probably we can keep the tests to catch these things in future, especially if we explicitly test the access count of those arrays after the reductions, to make sure we are not hitting fallbacks. We clearely are not doing that at the moment.

I am not sure about this we have these tests: https://github.com/meggart/DiskArrays.jl/blob/ccd3092ca6f32c2663ded1d74176aa9dd222a5ef/test/runtests.jl#L114 which also test the access count. Also the test that was added in this PR already passes the access_count test on the current main:

````julia
using DiskArrays
using DiskArrays.TestTypes
using Test
A = rand(1:10,30,30)
DA = AccessCountDiskArray(A,chunksize=(2,2));

r = sum(DA);

@test getindex_count(DA)==length(eachchunk(DA))
````

Here we test that every chunk is accessed exactly once, please try this on main, so I would guess the slow performance in #188 must come from somewhere else or some operation we are not yet hitting with our tests.

meggart:
Ok, I did some digging and realized that currently we do not hit our mapreduce methods but rather the iterator fallback, which does read the data chunk by chunk but is more inefficient and numerically unstable compared to this implementation. Will launch Cthulhu to find out where this goes wrong

rafaqz:
Seems this has been pretty productive! good to have all those things working and tested.

I think youre right - `sum`, `prod`, `any` and `all` are the real beneficiaries of not using `mapreduce`.  `maximum` and `minimum` are probably worth keeping too for reliability.


I don't quite understand how `count` is working over chunks here because it returns an `Int` but does not work on an iterator of `Int`. I thought it would need to be `sum` as the outer function and `count` as the inner.

meggart:
Superseded by #196

PR #193, opened 2024-09-19
Make AccessCountDiskArray <: AbstractDiskArray + fix zip bug

### LINKED ISSUE #175: `zip` triggers lots of invalidations
We have a workload over at [ClimaAtmos.jl](https://github.com/CliMA/ClimaAtmos.jl), where we use SnoopCompile's [invalidations script](https://timholy.github.io/SnoopCompile.jl/dev/tutorial/#Cut-to-the-Chase:-A-copy-paste-analysis-of-invalidations), and the result is:

```
[ Info: 4296 methods invalidated for 178 functions (showing 10 functions)
┌─────────────────────────────────────────────────────────┬───────────────┬───────────────┬─────────────────┐
│ <file name>:<line number>                               │ Function Name │ Invalidations │ Invalidations % │
│                                                         │               │               │     (xᵢ/∑x)     │
├─────────────────────────────────────────────────────────┼───────────────┼───────────────┼─────────────────┤
│ DiskArrays/bZBJE/src/zip.jl:83                          │      zip      │     3247      │       14        │
│ DiskArrays/bZBJE/src/zip.jl:70                          │      zip      │     3247      │       14        │
│ DiskArrays/bZBJE/src/zip.jl:83                          │      zip      │     3247      │       14        │
│ FillArrays/eOEVm/src/fillbroadcast.jl:78                │  broadcasted  │     1220      │        5        │
│ CommonDataModel/G3moc/src/groupby.jl:267                │  broadcasted  │     1220      │        5        │
│ ClimaCore/ANgUC/src/MatrixFields/field_name_dict.jl:331 │  broadcasted  │     1220      │        5        │
│ ClimaCore/ANgUC/src/MatrixFields/field_name_dict.jl:475 │  broadcasted  │     1220      │        5        │
│ ClimaCore/ANgUC/src/MatrixFields/lazy_operators.jl:44   │  broadcasted  │     1220      │        5        │
│ GeometryBasics/ebXl0/src/offsetintegers.jl:40           │    convert    │      834      │        4        │
│ DiskArrays/bZBJE/src/generator.jl:53                    │   Generator   │      439      │        2        │
└─────────────────────────────────────────────────────────┴───────────────┴───────────────┴─────────────────┘
```
So, DiskArrays' [`zip`](https://github.com/meggart/DiskArrays.jl/blob/64f4de86ec3ab921759d7372c19766a463c58e61/src/zip.jl#L81-L83) seems to invalidate 9741 methods.

charleskawczynski:
I'm hoping that this will alleviate or close #175.

### DIFF
--- a/src/util/testtypes.jl
+++ b/src/util/testtypes.jl
@@ -14,7 +14,7 @@ and optimise chunk access.
 `getindex_count(A)` and `setindex_count(A)` can be used to check the
 the counters.
 """
-struct AccessCountDiskArray{T,N,A<:AbstractArray{T,N},RS} <: AbstractArray{T,N}
+struct AccessCountDiskArray{T,N,A<:AbstractArray{T,N},RS} <: DiskArrays.AbstractDiskArray{T,N}
     getindex_log::Vector{Any}
     setindex_log::Vector{Any}
     parent::A
@@ -28,7 +28,7 @@ AccessCountDiskArray(a; chunksize=size(a),batchstrategy=DiskArrays.ChunkRead(Dis
 Base.size(a::AccessCountDiskArray) = size(a.parent)
 
 # Apply the all in one macro rather than inheriting
-DiskArrays.@implement_diskarray AccessCountDiskArray
+DiskArrays.@implement_diskarray_skip_zip AccessCountDiskArray
 
 DiskArrays.haschunks(::AccessCountDiskArray) = DiskArrays.Chunked()
 DiskArrays.eachchunk(a::AccessCountDiskArray) = DiskArrays.GridChunks(a, a.chunksize)
--- a/src/util/testtypes.jl
+++ b/src/util/testtypes.jl
@@ -28,7 +28,6 @@ AccessCountDiskArray(a; chunksize=size(a),batchstrategy=DiskArrays.ChunkRead(Dis
 Base.size(a::AccessCountDiskArray) = size(a.parent)
 
 # Apply the all in one macro rather than inheriting
-DiskArrays.@implement_diskarray_skip_zip AccessCountDiskArray
 
 DiskArrays.haschunks(::AccessCountDiskArray) = DiskArrays.Chunked()
 DiskArrays.eachchunk(a::AccessCountDiskArray) = DiskArrays.GridChunks(a, a.chunksize)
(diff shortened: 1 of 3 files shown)


charleskawczynski:
Oh, wait, @rafaqz, can we just move the TestTypes module outside of source? Why would DiskArrays ship that anyway?

felixcremer:
We use TestTypes in higher level packages  like YAXArrays to ensure that code is IO efficient.

charleskawczynski:
> We use TestTypes in higher level packages like YAXArrays to ensure that code is IO efficient.

Couldn't that just be shipped as a library instead?

felixcremer:
There might be a way to put this in an extension on Test, but then we would need to get the types out of the extension.

charleskawczynski:
> There might be a way to put this in an extension on Test, but then we would need to get the types out of the extension.

Yeah, an extension would work, or even just make a DiskArraysTestTypes package in this repo, and then just depend on that where needed. I don't think that this is a blocker for this PR, but I think moving it to a separate package is a cleaner solution entirely.

src/util/testtypes.jl
```diff
@@ -28,7 +28,7 @@ AccessCountDiskArray(a; chunksize=size(a),batchstrategy=DiskArrays.ChunkRead(Dis
 Base.size(a::AccessCountDiskArray) = size(a.parent)
 
 # Apply the all in one macro rather than inheriting
-DiskArrays.@implement_diskarray AccessCountDiskArray
+DiskArrays.@implement_diskarray_skip_zip AccessCountDiskArray
```
rafaqz, line 31:
```suggestion
```

This isn't needed, it's a subtype already

charleskawczynski: Ah, ok, thanks I'll update that

charleskawczynski:
I think I found a bug!

```julia
function Base.zip(A1::AbstractDiskArray, A2::AbstractArray, As::AbstractArray...)
    return DiskZip(A1, A1, As...)
end
```
should be
```julia
function Base.zip(A1::AbstractDiskArray, A2::AbstractArray, As::AbstractArray...)
    return DiskZip(A1, A2, As...)
end
```
! This is probably pretty confusing since it was correctly implemented in `implement_zip`!

rafaqz (approved):

PR #196, opened 2024-10-18
Update #189

meggart:
This is a rebased and slightly modified version of #189 . It keeps the optimized versions for some special reduction functions, and adding a test to check at least for `any` and `all` that early escape is happening.

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -47,3 +47,14 @@ macro implement_mapreduce(t)
         end
     end
 end
+
+
+# optimized implementation for special cases
+
+for fname in [:sum,:prod,:count,:all,:any,:minimum,:maximum]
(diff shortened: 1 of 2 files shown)


meggart:
@Alexander-Barth @rafaqz let me know what you think. Otherwise I would just merge...

src/mapreduce.jl
```diff
@@ -47,3 +47,20 @@ macro implement_mapreduce(t)
         end
     end
 end
+
+
+# Implementation for special cases for special cases and if fallback breaks in future julia versions
```
rafaqz, line 52:
```suggestion
# Implementation for special cases and if fallback breaks in future julia versions
```

rafaqz (approved):

Alexander-Barth:
Looks fine for me, Thanks!

PR #202, opened 2024-10-23
display warning (Can not determine size of element type...) only once

Alexander-Barth:
This PR makes sure that the warning 
```
Warning: Can not determine size of element type. Using DiskArrays.fallback_element_size[] = 100 bytes
```

is displayed only once. The warning is triggered quite often in NCDatastes' tests (for strings and variable-length arrays of different type).


meggart:
Thanks a lot. The repeated messages are indeed annoying.

PR #210, opened 2025-01-03
Allow constructing ConcatDiskArray from mixed type arrays

meggart:
and in addition improve performance by caching the chunks of the individual arrays instead of recomputing with each call to `eachindex`

### DIFF
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -4,12 +4,16 @@
 
 Joins multiple AbstractArrays or AbstractDiskArrays in lazy concatination.
 """
-struct ConcatDiskArray{T,N,P} <: AbstractDiskArray{T,N}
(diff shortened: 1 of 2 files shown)


src/cat.jl
```diff
@@ -51,12 +55,13 @@ function ConcatDiskArray(arrays::AbstractArray{<:AbstractArray{T,N},M}) where {T
     startinds = map(first, si)
     sizes = map(last, si)
 
-    return ConcatDiskArray{T,D,typeof(arrays1)}(arrays1, startinds, sizes)
+    chunks = concat_chunksize(D, arrays1)
+    hc = haschunks(first(arrays1))
```
rafaqz, line 59: Don't the separate concat arrays themselves count as chunks if there are no chunks internally?

meggart: Yes, that makes more sense. I updated and added a test for this

PR #219, opened 2025-01-27
add `PaddedDiskArray` and `pad`

rafaqz:
Ive had this use case a few times that you want to pad a disk array for windowing or fixing chunk alignment, and you don't want to make a copy of the file.

This PR adds a `PaddedDiskArray` wrapper and a `DiskArrays.pad` function that pads any disk array by custom amounts above and below the original array on each axis. The fill value for the padded area can be set with the `fill` keyword, which is zero by default.

Chunk offsets are updated to match the padding


src/pad.jl
```diff
@@ -0,0 +1,104 @@
...
+    PaddedDiskArray(A, padding, fill, chunks)
+end
+
+function _pad_offset(c::RegularChunks, (low, high), s)
+    chunksize = c.cs
+    offset = low > 0 ? c.offset - low + chunksize : c.offset
+    size = c.s + low + high
+    return RegularChunks(chunksize, offset, size)
+end
+function _pad_offset(c::IrregularChunks, (low, high), s)
```
meggart, line 38: It is not really obvious to me what the code should do here. Currently this adds an extra chunk at the beginning for the padded values. Should the values padded at the end be in a new chunk as well? Or should padded values simply be lumped into the first and last chunk. However, something is still wrong with the implementation, see this example:

````julia
c2 = IrregularChunks(chunksizes = [10, 10, 20, 30, 40])
_pad_offset(c2, (5,5), 100)
````

which returns chunks only up to 105 while the resulting array should have length 120.

File: src/pad.jl [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+
+Base.parent(A::PaddedDiskArray) = A.parent
+function Base.size(A::PaddedDiskArray)
+    map(size(parent(A)), A.padding) do s, (low, high)
+        s + low + high
+    end
+end
+
+readblock!(A::PaddedDiskArray, data, I...) = _readblock_padded!(A, data, I...)
+readblock!(A::PaddedDiskArray, data, I::AbstractVector...) =
```
meggart, line 59: Just curious why this is necessary, are the fallbacks too slow here?

src/pad.jl
```diff
@@ -0,0 +1,104 @@
...
+    fs = map(axes(parent(A)), Ipadded) do a, ip
+        searchsortedfirst(ip, first(a))
+    end
+    ls = map(axes(parent(A)), Ipadded) do a, ip
+        searchsortedlast(ip, last(a))
+    end
+    return if all(map(<=, fs, ls))
+        Idata = map(:, fs, ls)
+        Iparent = map(getindex, Ipadded, Idata)
+        @show Idata Iparent
```
meggart, line 80: should be removed

src/pad.jl
```diff
@@ -0,0 +1,104 @@
...
+haschunks(A::PaddedDiskArray) = haschunks(parent(A))
+eachchunk(A::PaddedDiskArray) = A.chunks
+
+function _readblock_padded(A, data, I...)
+    data .= A.fill
+    Ipadded = map(I, A.padding) do i, (low, high)
+        i .- low
+    end
+    fs = map(axes(parent(A)), Ipadded) do a, ip
+        searchsortedfirst(ip, first(a))
```
meggart, line 72: This assumes indices `I` are sorted, so this readblock implementation should really only be called for I::AbstractUnitRange and not for `AbstractVector`

test/runtests.jl
```diff
@@ -962,6 +962,29 @@ end
...
+    regions = pa[10:20, 30:40, 1:2]
+    testdata = copy(regions) .= 999
+    testdata[2:11, 2:11, 2:2] .= ch[1:10, 1:10, 1:1]
+    @test regions == testdata
+    # Overlapping upper
+    regions = pa[101:120, 141:160, 5:7]
+    testdata = copy(regions) .= 999
+    testdata[1:10, 1:10, 1:1] .= ch[91:100, 111:120, 4:4]
+    testdata
+    @test regions == testdata
```
meggart, line 985: Maybe add the following tests for _pad_offset:

````julia
c1 = RegularChunks(10, 0, 100)
_pad_offset(c1, (5,2), 100) == RegularChunks(10,5,107)
_pad_offset(c1, (30,2), 100) == RegularChunks(10,0,132)

c2 = IrregularChunks(chunksizes = [10, 10, 20, 30, 40])
#The following test would assume padding ends up in a separate chunk:
_pad_offset(c2, (5,5), 100) == IrregularChunks(chunksizes = [5,10, 10, 20, 30, 40,5])
````

**Review by meggart — CHANGES_REQUESTED** [review]
Thanks for starting this, this would be useful indeed.

**Comment by meggart** [comment]
@rafaqz do you need support for fixing some of the chunk issues?

**Comment by rafaqz** [comment]
Should be ok, just haven't had time to get back to it yet

File: src/pad.jl [hunk 6]
```diff
@@ -0,0 +1,104 @@
...
+function PaddedDiskArray(A::AbstractArray{T,N}, padding::NTuple{N,Tuple{Int,Int}};
+    fill=zero(eltype(A)),
+) where {T,N}
+    chunks = GridChunks(map(_pad_offset, eachchunk(A).chunks, padding, size(A)))
+    PaddedDiskArray(A, padding, fill, chunks)
+end
+
+function _pad_offset(c::RegularChunks, (low, high), s)
+    chunksize = c.cs
+    offset = low > 0 ? c.offset - low + chunksize : c.offset
```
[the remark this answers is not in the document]
rafaqz: Ok I've added a fix for this for arbitrary padding sizes

src/pad.jl
rafaqz: Thanks, IrregularChunks needs more work

test/runtests.jl
rafaqz: Ok I'v added these

src/pad.jl
rafaqz: fixed

rafaqz:
@meggart I think this is good to go

meggart (approved):

PR #220, opened 2025-01-30
allow empty chunks

rafaqz:
This needs tests, but wonddering @meggart is there a reason this isn't allowed? It seems to come up sometimes that arrays end up empty and filtering those cases out is more annoying than just handling the empty case

### DIFF
--- a/src/chunks.jl
+++ b/src/chunks.jl
@@ -23,8 +23,9 @@ struct RegularChunks <: ChunkType
     offset::Int
     s::Int
     function RegularChunks(cs::Integer, offset::Integer, s::Integer)
-        cs>0 || throw(ArgumentError("Chunk sizes must be strictly positive"))
-        -1 < offset < cs || throw(ArgumentError("Offsets must be positive and smaller than the chunk size"))
+        cs >= 0 || throw(ArgumentError("Chunk sizes must be strictly positive"))
+        (cs == 0 && offset == 0) || (-1 < offset < cs) || 
+            throw(ArgumentError("Offsets must be positive and smaller than the chunk size"))
         s >= 0 || throw(ArgumentError("Negative dimension lengths are not allowed"))
         new(Int(cs), Int(offset), Int(s))
     end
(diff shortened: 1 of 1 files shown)


meggart:
I don't think empty arrays are a problem. It is just a zero chunk size that IMO does not make a lot of sense. I have not been following issues in the last week, is there an example where this came up?

rafaqz:
I will have to investigate, but somehow I'm hitting this empty error with regular code, but it's deeply nested so not easy to demo.

Maybe it needs to be caught further out, I can try that approach.

meggart:
Yes, this would be good to investigate. Usually this happens when there are some fallback definitions for `eachchunk` for custom array types that go like `RegularChunks(s,0,s)` where `s` is the length along the dimension, which should rather be `RegularChunks(max(s,1),0,s)`. I wonder if we could modify the inner constructor of `RegularChunks` slightly to something like this:

````julia
struct RegularChunks <: ChunkType
  cs::Int
  offset::Int
  s::Int
  function RegularChunks(cs::Integer, offset::Integer, s::Integer)
      if cs < 0 
        throw(ArgumentError("Chunk sizes must be strictly positive"))
      elseif cs == 0
        if s == 0
          #This is a special case we allow and where we set the chunk size to 1
          cs = 1
        else
          throw(ArgumentError("Chunk sizes must be strictly positive"))
        end
      end
      -1 < offset < cs || throw(ArgumentError("Offsets must be positive and smaller than the chunk size"))
      s >= 0 || throw(ArgumentError("Negative dimension lengths are not allowed"))
      new(Int(cs), Int(offset), Int(s))
  end
end
````

So we repair the special case when both chunk size and dimension length are 0. No idea what we should do here, but would definitely be good to find out where the zero-chunksize is coming from. I can investigate as well, even if it is nested, would just be good to have a pointer. I think I have seen this error when creating empty slice views into Rasters, but can not remember details.

meggart:
I would suggest to leave this PR as stale for now and see if problems keep occurring after #232 . If problems are still reported, lets discuss this again.

meggart:
Looks like there are no urgent issues for no, so I would close. Feel free to re-open if necessary

PR #221, opened 2025-02-01
bugfix and improve `isdisk`

rafaqz:
The macro was not applying `isdisk` properly, and it really should be on the type.

### DIFF
--- a/src/diskarray.jl
+++ b/src/diskarray.jl
@@ -13,8 +13,9 @@ abstract type AbstractDiskArray{T,N} <: AbstractArray{T,N} end
 Return `true` if `a` is a `AbstractDiskArray` or follows 
 the DiskArrays.jl interface via macros. Otherwise `false`.
 """
-isdisk(a::AbstractDiskArray) = true
-isdisk(a::AbstractArray) = false
+isdisk(a::AbstractDiskArray) = isdisk(typeof(a))
+isdisk(::Type{<:AbstractDiskArray}) = true
+isdisk(::Type{<:AbstractArray}) = false
 
 """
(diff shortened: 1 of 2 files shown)


meggart (approved):

src/diskarray.jl
```diff
@@ -13,8 +13,9 @@ abstract type AbstractDiskArray{T,N} <: AbstractArray{T,N} end
 Return `true` if `a` is a `AbstractDiskArray` or follows 
 the DiskArrays.jl interface via macros. Otherwise `false`.
 """
-isdisk(a::AbstractDiskArray) = true
-isdisk(a::AbstractArray) = false
+isdisk(a::AbstractDiskArray) = isdisk(typeof(a))
```
rafaqz, line 16:
```suggestion
isdisk(a::AbstractArray) = isdisk(typeof(a))
```

tiemvanderdeure:
I think this causes some methods to be overwritten during precompilation though

```
┌ DiskArrays [3c3547ce-8d99-4f5e-a174-61eb10b00ae3]
│  WARNING: Method definition isdisk(Type{var"#s25"} where var"#s25"<:(DiskArrays.AbstractDiskArray{T, N} where N where T)) in module DiskArrays at /home/runner/work/DiskArrays.jl/DiskArrays.jl/src/diskarray.jl:17 overwritten at /home/runner/work/DiskArrays.jl/DiskArrays.jl/src/diskarray.jl:380.
│    ** incremental compilation may be fatally broken for this module **
```

tiemvanderdeure:
Those things are so easy to overlook. Would be nice if you could make it error

PR #224, opened 2025-02-05
fast path for common_chunks where only one disk array

### LINKED ISSUE #222: `common_chunks` on `SubDiskArray` can be a large fraction of io cost
Image

This is where the left hand side of the broadcast is a `Matrix` and the right is a `SubDiskArray`

Probably `common_chunks` should never be called for this case?

(for context this is completely killing the performance reading thousands of 500*500 windows from a large lazy disk array into a preallocated matrix)

rafaqz:
Closes #222

At least for the case of a single disk array, the profile is back to being all GDAL reads.

However it can still be slow for multiple GDAL style column chunked arrays.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -60,26 +60,24 @@ function common_chunks(s, args...)
     if isempty(chunkedarrays)
         totalsize = sum(sizeof ∘ eltype, args)
         return estimate_chunksize(s, totalsize)
+    elseif length(chunkedarrays) == 1
+        return eachchunk(only(chunkedarrays))
     else
-        if length(chunkedarrays) > 1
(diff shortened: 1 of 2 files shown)


rafaqz:
Maybe we need to define == on chunks so it doesn't iterate over all the values?

rafaqz:
Ok I added `==` on chunks, its a little complicated for RegularChunks as the chunks can match with different chunk size and offset for one or two chunks.

But should help performance of `common_chunks` in the multiple array case as well.

rafaqz:
@meggart can you have a final look at this? I added equality on chunks

meggart:
Looks good!

PR #225, opened 2025-02-06
drop isdisk definition on AbstractDiskArray

tiemvanderdeure:
Reverse a bug introduced by https://github.com/JuliaIO/DiskArrays.jl/pull/221. `isdisk` is defined twice for AbstractDiskArray (directly and because the implement_getindex macro is called), which affects precompilation

### DIFF
--- a/src/diskarray.jl
+++ b/src/diskarray.jl
@@ -14,7 +14,6 @@ Return `true` if `a` is a `AbstractDiskArray` or follows
 the DiskArrays.jl interface via macros. Otherwise `false`.
 """
 isdisk(a::AbstractArray) = isdisk(typeof(a))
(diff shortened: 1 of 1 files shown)


rafaqz:
Yep, forgot about that

tiemvanderdeure:
Does Aqua have anything to test for this? Maybe test_persistent_tasks? https://juliatesting.github.io/Aqua.jl/stable/persistent_tasks/#Aqua.test_persistent_tasks

tiemvanderdeure:
the test failures are because some server is down

rafaqz:
Not sure about the tests, but I'll release this ASAP this will mess with a lot of people

rafaqz:
Oh it was only from the second pr. Well, version is bumped anyway ..

PR #226, opened 2025-02-06
all `Chunked`s are chunked

tiemvanderdeure:
all `Chunked`s are equal!

I noticed something like a raster doesn't necessary use the default `Chunked()`, which means it wouldn't recognised as a chunked array by `common_chunks`, and lose it's chunking e.g. in this case

```julia
using Rasters, DiskArrays, NCDatasets
ras1 = Rasters.create("ras1.nc", Float64, (X(1:100), Y(1:100)); force = true, chunks = (10, 10))
DiskArrays.eachchunk(ras1 .+ 1)
```

I want to add a test but don't none of the arrays that are tested seem to have other chunks than `Chunked()` - anyone who understands this more could maybe chip in?


rafaqz:
Oof this will clash with my other common_chunks PR, but yes that makes more sense as Chunked isn't a singleton now. 

Is there a test for where this would break?

tiemvanderdeure:
Ah I didn't see your PR. Feel free to just take the fix and I'll close this PR (unless you can do some git magic and I get my hard-earned commit authorship?)

rafaqz:
We can merge this and do it there - but mostly I don't want to take responsibility for this PR because I don't have time to write the tests

meggart (approved):

rafaqz:
Done

PR #229, opened 2025-02-08
refactor, rename and document

### LINKED ISSUE #223: Heavily abbreviated code may be blocking contributions
A lot of the code here is pretty hard to understand simply because its a complicated problem we are solving.

But also a lot of it uses abbreviated variable names and struct fields so that outsiders (including me for half of it!) don't easily understand whats happening, before even getting to how complicated the problem is.

Spelling everything out a bit more may help contributors. For example, `cs` could be `chunksize` and `chunkedars` `chunkedarrays`.

rafaqz:
This PR renames variables and functions to be more explicit, reorganises files, and adds a lot more documentation and code comments.

The aim is to make the interface easier to understand and implement for developers, and make contributing here a little easier.

Closes #223


src/chunks.jl
```diff
@@ -1,272 +1,295 @@
...
-        #All chunks have the same size, only first and last chunk can be shorter
-        chunksize = chunks[2]
-        return RegularChunks(chunksize, chunksize - chunks[1], sum(chunks))
+# Utility function that constructs either a `RegularChunks` or an
+# `IrregularChunks` object based on a vector of chunk sizes given as worted Integers. Wherever
+# possible it will try to create a regular chunks object.  
+function chunktype_from_chunksizes(chunksizes::AbstractVector)
+    if length(chunksizes) == 1
+        # only a single chunk is affected
+        return RegularChunks(chunksizes[1], 0, chunksizes[1])
```
meggart, line 233: Should be fixed here or in a separate PR: fix the case where `chunksizes[1] == 0` maybe this is even the culprit of some of the errors you are seeing.

rafaqz: Maybe separate PR I'm trying not to change any logic here, it's just renames etc

PR #230, opened 2025-02-10
Fix bug in testtypes

meggart:


meggart:
@rafaqz I just realized how big #229 is. So maybe it is better to merge that one first and then I re-apply this fix afterwards.

rafaqz:
Yeah we can hope it missed those lines! But yes good to merge it soon

PR #234, opened 2025-02-13
make `cat_disk` run on a vector rather than a vararg

asinghvi17:
Ideally this should allow us to run `cat` across a large number of arrays.

### DIFF
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -140,7 +140,7 @@ mergechunks(a::ChunkVector, b::ChunkVector) = mergechunks_irregular(a, b)
 mergechunks_irregular(a, b) =
     IrregularChunks(; chunksizes=filter(!iszero, [length.(a); length.(b)]))
 
-function cat_disk(dims, As::AbstractArray...)
+function cat_disk(dims, As::AbstractVector{<: AbstractArray})
     if length(dims) == 1
         dims = only(dims)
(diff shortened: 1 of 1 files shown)


felixcremer:
Shouldn't we rather implement Base.stack instead? From the docstring below it seems to me, that this is the function we aim for when we want to concatenate a list of DiskArrays:

```julia
  stack(iter; [dims])

  Combine a collection of arrays (or other iterable objects) of equal size into one larger array, by arranging them
  along one or more new dimensions.
```

asinghvi17:
The key word there is _new_ dimensions...the problem here was to cat along a preexisting one

rafaqz:
I don't get what's achieved here, there's already a splat into the base function.

If your going to call a DiskArrays method anyway `ConcatDiskArray` already works how you want, even on multidimensional arrays, I suggest doing that. `cat_disk` is literally just `cat` for disk arrays, and it's internal. It's just separate so it's not stuck inside the macro

asinghvi17:
Yeah, I had wondered if this might speed up the compilation a bit so tried it on my phone.  Tested it and apparently try not :D

rafaqz:
Where we could do this is with a reduce vcat hack that Base also has...

asinghvi17:
Should we just use that method for combine on a stack as well then,

Issue #35, opened 2021-06-05
 Chunks do not align in dimension 1


rafaqz:
Can we work around this? This being an error is a problem for using chunked broadcast as a default.

I'm not totally across the underlying mechanisms yet, but I can imagine 2 possibilities: 
- force chunk alignment by loading multiple chunks to a temporary chunk of the right size
- load the whole array to memory and ignore chunking, especially if the array is small

Then the error could be a warning.

meggart:
There are certainly ways to work around this, but I don't know how to do it with the current broadcast syntax. Ideally one would force the broadcast operation to use the chunks of the array with the highest access latency, and usually the user should know which one this is. We could say that in case where chunks do not align, we just give preference to the first array in the broadcast operation and document that users should always do: `largearray .+ smallarray` instead of `smallarray .+ largearray`?

felixcremer:
Could we also get a setchunks function like we have in YAXArrays ? So that one could be very explicit about defining which arrays to rechunk for the broadcast to work.

I tried looking into this problem, but I don't really understand how I would have to change the inner code to make it work more flexible.

rafaqz:
We have RenchunkedDiskArray here, do we just need a `rechunk` function?

felixcremer:
Yes, then we could have a rechunk method for DimArrays as well.

rafaqz:
Ok, if you add `rechunk` here I can extend it in DD

meggart:
This is resolved. Misaligned broadcast works via fallback, and MockChunkedDiskArray allows rechunking for explicit alignment control through mockchunks functions. Alternatively, users can use DiskArrayEngine backend once #316 is merged if they do not want to care about manual chunks alignment.

PR #237, opened 2025-02-27
Add mockchunks function

felixcremer:
This adds a mockchunks function which would return a MockChunkedDiskArray
This function could then be overloaded by wrapping packages like DimensionalData. 

This was discussed in #35 but is not a fix to that but would open up a better workaround. 
This would need some tests


src/rechunk.jl
```diff
@@ -14,6 +14,13 @@ struct RechunkedDiskArray{T,N,A<:AbstractArray{T,N},C<:GridChunks} <: AbstractDi
     chunks::C
 end
 
+"""
+    rechunk(A::AbstractArray, chunks)
+Rechunk the underlying data of A into the given `chunks`.
+
+"""
+rechunk(data::AbstractDiskArray, chunks::GridChunks) = RechunkedDiskArray(data , chunks)
```
asinghvi17, line 22:
```sugggestion
rechunk(data::AbstractDiskArray, chunks::GridChunks) = RechunkedDiskArray(data, chunks)
```

just for style

asinghvi17 (commented):
Would it make sense to allow any sensible definition of `chunks` here?  I think rechunkeddiskarray takes tuples of chunksize like `(10, 10, 3)` for instance...

asinghvi17 (commented):
Would it make sense to allow any sensible definition of `chunks` here?  I think rechunkeddiskarray takes tuples of chunksize like `(10, 10, 3)` for instance...

meggart (approved):

meggart:
I start wondering if `rechunk` might be a misleading function name. In the end, we are not rechunking the data but rather creating a view into the data that behaves as if it was chunked in a different way internally. There might be some confusion among users when they assume that the result of `rechunk` would magically make time series access more efficient on spatially chunked data, this happens quite reularly in the python world for xarray's `chunk` function. 

I think we need either a different name or at least a better docstring explaining that no physical rechunking is happening here.

felixcremer:
Yes, i am also wondering whether the name setchunks makes more sense. I could copy the docstring from the YAXArrays.setchunks function and remove the YAXArray specifics. 
```
help?> YAXArrays.setchunks
  setchunks(c::YAXArray,chunks)

  Resets the chunks of a YAXArray and returns a new YAXArray. Note that this will not change the chunking of the underlying data itself, it will just make the data
  "look" like it had a different chunking. If you need a persistent on-disk representation of this chunking, use savecube on the resulting array. The chunks argument can
  take one of the following forms:

    •  a DiskArrays.GridChunks object

    •  a tuple specifying the chunk size along each dimension

    •  an AbstractDict or NamedTuple mapping one or more axis names to chunk sizes
```

rafaqz:
Even `setchunks seems more active... It's really like `fakechunks`

felixcremer:
Fakechunks sounds like it is not chunked at all. But I see your point. 
Other possible names could be:

pretendchunk
pseudochunk
mockchunk

And if we rename the function we should also rename the main object because it would be good to have them in accordance.

meggart:
Is there any conclusion/progress on this one. It would be a shame if this got blocked just because we can't agree on a name. So, maybe we discuss in our next big Raser call?

rafaqz:
Can we use `mockchunks` and `MockChunkedArray` and get this merged?

rafaqz (approved):

rafaqz:
@meggart if you're happy with the new name this looks good to merge to me

rafaqz:
Oh one thing @felixcremer can we add `const RechunkedDiskArray = MockChunksDiskArray` or whatever it is so this isn't a breaking change?

Issue #242, opened 2025-03-11
unique and count


bjarthur:
next question:   DiskArrays.jl seems like GPUArrays.jl to me in that it implements various Base functions optimized for a special subtype of AbstractArray.  if that's the case, should i submit a PR for these methods?

```
Base.unique(a::AbstractDiskArray) = union((unique(a[c...]) for c in eachchunk(a))...)
Base.count(a::AbstractDiskArray) = sum(count(a[c...]) for c in eachchunk(a))
```

rafaqz:
Yes, very similar to GPUArrays like that.

We have `count` already here:

https://github.com/JuliaIO/DiskArrays.jl/blob/main/src%2Fmapreduce.jl#L63

`unique` probably helps, although only if there are not too many repeated values? or it will run out of memory? Probably it needs to check memory at the start then run in a `for` loop and update the total as it goes, and errors if it's running out.

Are there other methods like that? We did sum, prod, minimum, maximum, count last time this came up.

bjarthur:
aha, my code wasn't dispatching to `DiskArray.count` as i was using the 1-argument method.  would be good to add `count(v::AbstractDiskArray) = count(identity, v::AbstractDiskArray)`.

bjarthur:
did you mean to say too many *unique* values instead of "repeated"?  otherwise, i don't understand how it could OOM.  does `Base.unique` worry about memory.  if not, why should DiskArrays.jl?

rafaqz:
Yes like with some Float64 data you would expect the unique data to be in the same order of size as the original data. A 200GB file might 50 GB unique, and there goes your RAM on many systems. So if we run `unique` chunked we may as well stop that situation with an error rather than crashing julia.

bjarthur:
could just limit it to `unique(::AbstractDiskArray{Int}`, but i'd be inclined to let the user do what they want and it's on them in it OOMs

rafaqz:
Fair enough, and best not to limit it by type.

rafaqz:
This is implemented now

PR #243, opened 2025-03-17
add 1-arg count, sum, prod, etc. methods; with tests

bjarthur:
as [discussed](https://github.com/JuliaIO/DiskArrays.jl/issues/242#issuecomment-2717960630)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,13 +14,17 @@ LRUCache = "1"
 Mmap = "1"
 OffsetArrays = "1"
 Statistics = "1.9"
+Suppressor = "0.2.8"
 Test = "1.9"
+TraceFuns = "0.2.0"
 julia = "1.9"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
+TraceFuns = "626f57eb-f6ca-464c-8f04-384fee2a352e"
 
 [targets]
-test = ["Aqua", "Statistics", "Test"]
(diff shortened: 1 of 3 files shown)


bjarthur:
tests pass.  ready for review

bjarthur:
hmm...  should probably handle the dims and init kwargs while were at it.

bjarthur:
ok, so, adding only the init kwarg to count caused the tests to fail [here](https://github.com/JuliaIO/DiskArrays.jl/blob/main/test/runtests.jl#L144).  must be some sort of dispatch problem since that line tries to pass in a dims kwarg.

and adding a dims kwarg is more than i can chew a.t.m, particularly given that one can simply use the dims and init kwargs of mapreduce instead.

so i reverted the init kwarg commit, and added *one*-arg (hah, not 2, misspoke in the original title to this PR) methods to sum, prod, etc. just above.

rafaqz (approved):

rafaqz:
Yeah, dims keywords are HARD! 

We could add methods for `sum` `mean` `maximim` etc but not for anonymous funcions or `median`.

bjarthur:
> We could add methods for `sum` `mean` `maximim` etc but not for anonymous funcions or `median`.

can you please clarify?  i don't understand.

bjarthur:
access counts are the same i now understand because without this PR the fallback is to the [Generator](https://github.com/JuliaIO/DiskArrays.jl/blob/d62bf4915731542410bb15f40f34b169a6fab202/src/generator.jl#L9-L12).

rafaqz:
> We could add methods for `sum` `mean` `maximim` etc but not for anonymous functions or `median`

I mean we know order doesn't matter with these methods, so we can allocate the final array size then reduce each chunk into it in the right places.

But generally we don't know that so it would be specialised.

bjarthur:
i figured out a way to test that the new methods do not use the generator.  involves tracing the called functions, and requires two extra test dependancies.  let me know what you think.

rafaqz:
That could be generally useful to have for other tests too. @meggart what do you think about test deps

meggart (approved):

Issue #246, opened 2025-03-18
mapreduce fallback tracking issue


bjarthur:
in trying to understand why [`count` was slow](https://github.com/JuliaIO/DiskArrays.jl/issues/242), i came across a comment explaining that it [used to fallback to `mapreduce`](https://github.com/JuliaIO/DiskArrays.jl/pull/189#issuecomment-2315712804) but now does not because of internal refactorings of the API.  so i dug through this a bit so as to learn how DiskArrays worked, and found that the [Generator](https://github.com/JuliaIO/DiskArrays.jl/blob/d62bf4915731542410bb15f40f34b169a6fab202/src/generator.jl#L9-L12) is hit instead of [`mapfoldl_impl`](https://github.com/JuliaIO/DiskArrays.jl/blob/main/src/mapreduce.jl#L28-L47) at least back to julia v1.9.  didn't go back further as that is the current minimum version required by DiskArrays.jl.

given enough developer time, it'd be nice to have these fallback method work again.  in the meantime, we might consider [blocking them out with `VERSION`](https://github.com/JuliaGPU/CUDA.jl/blob/d223d48d8231b27987bd5eab86663f80b6f90396/src/pointer.jl#L76-L84) as is commonly done in CUDA.jl.  probably shouldn't delete them entirely (yet), as they would serve as a template for updating them.

PR #247, opened 2025-03-24
add dims kwarg

bjarthur:
not battle tested yet.  and might as well add an `init` kwarg too

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -79,7 +79,7 @@ for (_fname, init, acum) in ((:_sum, :zero, :+), (:_prod, :one, :*),
 		_dims = typeof(dims)<:Tuple ? [dims...] : typeof(dims)<:Number ? [dims] : dims
 		out_dims = [size(a)...]
 		out_dims[_dims] .= 1
-		out = fill($init(eltype(a)), out_dims...)
+		out = fill($init(Base.return_types(f, (eltype(a),))[1]), out_dims...)
 		for c in eachchunk(a)
 			out_c = [c...]
 			out_c[_dims] .= Ref(1:1)
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -80,7 +80,8 @@ for (_fname, init, acum) in ((:_sum, :zero, :+), (:_prod, :one, :*),
 		out_dims = ntuple(Val(N)) do i
 			i in _dims ? 1 : size(a)[i]
 		end 
-		out = fill($init(Base.return_types(f, (T,))[1]), out_dims...)
+		T1 = Base.promote_op(f, T)
+		out = fill($init(T1), out_dims...)
 		for c in eachchunk(a)
 			out_c = ntuple(Val(N)) do i
 				i in _dims ? (1:1) : c[i]
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -51,9 +51,13 @@ end
 
 # Implementation for special cases and if fallback breaks in future julia versions
 
-for fname in [:sum, :prod, :all, :any, :minimum, :maximum]
-    @eval Base.$fname(v::AbstractDiskArray) = Base.$fname(identity, v::AbstractDiskArray)
-    @eval function Base.$fname(f::Function, v::AbstractDiskArray)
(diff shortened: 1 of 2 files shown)


src/mapreduce.jl
```diff
@@ -66,3 +70,21 @@ function Base.count(f, v::AbstractDiskArray)
...
 end
+
+for (_fname, init, acum) in ((:_sum, :zero, :+), (:_prod, :one, :*),
+                             (:_all, _->:true, :&), (:_any, _->:false, :|),
+                             (:_maximum, :typemin, :max), (:_minimum, :typemax, :min))
+	@eval Base.$_fname(a::AbstractDiskArray, dims)  = Base.$_fname(identity, a, dims)
+	@eval function Base.$_fname(f::Function, a::AbstractDiskArray, dims)
+		_dims = typeof(dims)<:Tuple ? [dims...] : typeof(dims)<:Number ? [dims] : dims
+		out_dims = [size(a)...]
+		out_dims[_dims] .= 1
```
rafaqz, line 81:
```julia
out_dims = ntuple(Val{N}) do i
    i in _dims ? 1 : size(a)[i]
end 
```

Something like this should be the same but type stable for splats. `N` needs a `where N` on the array ndims type parameter.

bjarthur: added a commit with this code.  the `out` array is still type unstable.  not sure how to fix that given it's eltype is calculated.

bjarthur:
also need to search for something other than "DiskGenerator" in the output of `@trace` as that is not present on main either.

src/mapreduce.jl
```diff
@@ -66,3 +70,23 @@ function Base.count(f, v::AbstractDiskArray)
...
+for (_fname, init, acum) in ((:_sum, :zero, :+), (:_prod, :one, :*),
+                             (:_all, _->:true, :&), (:_any, _->:false, :|),
+                             (:_maximum, :typemin, :max), (:_minimum, :typemax, :min))
+	@eval Base.$_fname(a::AbstractDiskArray, dims)  = Base.$_fname(identity, a, dims)
+	@eval function Base.$_fname(f::Function, a::AbstractDiskArray{T,N}, dims) where {T,N}
+		_dims = typeof(dims)<:Tuple ? [dims...] : typeof(dims)<:Number ? [dims] : dims
+		out_dims = ntuple(Val(N)) do i
+			i in _dims ? 1 : size(a)[i]
+		end 
+		out = fill($init(Base.return_types(f, (T,))[1]), out_dims...)
```
rafaqz, line 83:
```suggestion
                T1 = Base.promote_op(f, T)
		out = fill($init(T1), out_dims...)
```

I think this is more likely to be stable? (untested)


Isn't this the syntax? `(T,)` wont be type stable.

meggart:
Why exactly do we need these special implementations? Has something in the Julia mapreduce internals changed so that all of these don't hit the `mapreducedim!` implementation anymore. See also this PR for a similar discussion.

bjarthur:
> Why exactly do we need these special implementations? Has something in the Julia mapreduce internals changed so that all of these don't hit the `mapreducedim!` implementation anymore. See also this PR for a similar discussion.

which PR?  and no, nothing falls back to mapreduce now as the internals have changed.  https://github.com/JuliaIO/DiskArrays.jl/issues/246

rafaqz:
Right yes we have this code already:

https://github.com/JuliaIO/DiskArrays.jl/blob/main/src/mapreduce.jl#L13-L26

Probably we need to hook this up againg? It could also do with some comments and cleanup so its clear what its all for

bjarthur:
oops, i take that back.  `maximum` is indeed falling back to https://github.com/JuliaIO/DiskArrays.jl/blob/main/src/mapreduce.jl#L13, it's `count` that isn't falling back to https://github.com/JuliaIO/DiskArrays.jl/blob/main/src/mapreduce.jl#L28.  quick benchmarks show that this PR is no faster.  shoot.

meggart:
Is this PR still active?

bjarthur:
i think it can be closed.

but has mapreduce been updated?  https://github.com/JuliaIO/DiskArrays.jl/issues/246

PR #249, opened 2025-04-01
add PermutedDimsArray method that returns PermutedDiskArray

### LINKED ISSUE #240: Handle `PermutedDimsArray`
The chunks of a `PermutedDimsArray` of a chunked array is not chunked - this is really annoying for Rasters in particular, because the `@d` macro internally uses `PermutedDimsArray`

```julia
using Rasters, NCDatasets
import Rasters.DiskArrays: eachchunk
A = Rasters.create(
    "myraster.nc", Float64, (X(1:1000), Y(1:1000)); 
    force = true, chunks = (128, 128)
)

Aperm = @d A (Y, X)
eachchunk(Aperm) # 1×1 DiskArrays.GridChunks
eachchunk(A) # 8x8 DiskArrays.GridChunks
```

Or a more minimal example

```
using DiskArrays
chunkedA = DiskArrays.TestTypes.ChunkedDiskArray(rand(100,100), (10, 10))
PermutedDimsArray(chunkedA, (2,1)) |> eachchunk
```
```
1×1 DiskArrays.GridChunks{2, Tuple{DiskArrays.RegularChunks, DiskArrays.RegularChunks}}:
 (1:100, 1:100)
```

I'm not entirely sure how this should work - probably we need to get the chunks of the parent and then apply the same permutation, right?

rafaqz:
Closes #240

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,12 +4,14 @@ authors = ["Fabian Gans <fgans@bgc-jena.mpg.de>"]
 version = "0.4.11"
 
 [deps]
+ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
 Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 
 [compat]
 Aqua = "0.8"
+ConstructionBase = "1"
 LRUCache = "1"
 Mmap = "1"
 OffsetArrays = "1"
--- a/src/DiskArrays.jl
+++ b/src/DiskArrays.jl
@@ -1,5 +1,7 @@
 module DiskArrays
 
+import ConstructionBase
+
 using LRUCache: LRUCache, LRU
 
 # Use the README as the module docs
--- a/src/permute.jl
+++ b/src/permute.jl
@@ -4,20 +4,31 @@
 A lazily permuted disk array returned by `permutedims(diskarray, permutation)`.
 """
 struct PermutedDiskArray{T,N,perm,iperm,A<:AbstractArray{T,N}} <: AbstractDiskArray{T,N}
-    a::A
+    parent::A
 end
+# We use PermutedDimsArray internals instead of duplicating them,
+# and just copy the type parameters it calculates.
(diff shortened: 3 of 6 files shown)


rafaqz:
Ok so to remove the recursive PermuteDimsArray I moved type parameters to PermutedDiskArray. But then I had to add a ConstructionBase dependency so we can use Flatten.jl/Accessors.jl etc on it (so Rasters and other things don't break). ConstructionBase.jl is tiny will nearly always be loaded anyway.

Then there were some issues with parent access after changes to PermutedDiskArray, so I went ahead and swapped everything to use `parent(a)`.

meggart (approved):
Looks good to me. Would you consider this change to be breaking?

rafaqz:
I guess it could be considered technically breaking, But I imagine the scope of the break is fairly limited. The `PermutedDimsArray` was essentially useless anyway as it was scalar indexing, so its not likely to break any real workflows.

tiemvanderdeure:
I guess DD is going to require this to make permutedims work, right? So then it really wouldn't be ideal to release a 0.5 - it would cause compatibility issues with the whole ecosystem until everyone updates.

A `PermutedDimsArray` of a DiskArray was basically unusable anyway.

rafaqz:
Absolutely, I think we should just st bump a patch version

mtfishman:
I hadn't noticed this change until now, but this broke some code in one of my packages where I define my own `AbstractDiskArray` subtype, and I have code that assumed that `PermuteDimsArray(::MyDiskArray, perm)` returns a `PermutedDimsArray` (which I then convert to me own `PermutedDiskArray`-like type, similar to the previous logic in DiskArrays.jl). While I understand the motivation behind it, I think this behavior is pretty surprising, what about defining:
```julia
permuteddims(a::AbstractArray, perm) = PermutedDimsArray(a, perm)
permuteddims(a::AbstractDiskArray, perm) = PermutedDiskArray(a, perm)
```
and then use `permuteddims` inside of DiskArrays.jl in places where you might return a `PermutedDiskArray`? That's what I do in some of my own packages. Of course it would be ideal if that kind of design was in Base, though in the meantime we could make a small interface package that defines `permuteddims` that is shared across packages.

rafaqz:
The problem is the interface to other packages and users that don't know about this method. And that anyone ever calling PermuteDimsArrray on a disk array is fundamentally broken (it causes scalar indexing) and shouldnt be allowed to happen in any case.


We really need a method in Base Julia, as you have read in my issue.

PR #255, opened 2025-04-10
add abstract types for views and permuted arrays

tiemvanderdeure:
Per @lupemba 's excellent suggesion in https://github.com/JuliaGeo/NCDatasets.jl/issues/274#issuecomment-2785344447, this PR implements an AbstractPermutedDiskArray and AbstractSubDiskArray

### DIFF
--- a/src/permute.jl
+++ b/src/permute.jl
@@ -3,14 +3,14 @@
 
 Abstract supertype for diskarray with permuted dimensions.
 """
-abstract type AbstractPermutedDiskArray{T,N} <: AbstractDiskArray{T,N} end
+abstract type AbstractPermutedDiskArray{T,N,perm,iperm,A} <: AbstractDiskArray{T,N} end
 
 """
     PermutedDiskArray <: AbstractPermutedDiskArray
 
 A lazily permuted disk array returned by `permutedims(diskarray, permutation)`.
 """
-struct PermutedDiskArray{T,N,perm,iperm,A<:AbstractArray{T,N}} <: AbstractPermutedDiskArray{T,N}
+struct PermutedDiskArray{T,N,perm,iperm,A<:AbstractArray{T,N}} <: AbstractPermutedDiskArray{T,N,perm,iperm,A}
     parent::A
 end
 # We use PermutedDimsArray internals instead of duplicating them,
@@ -64,8 +64,8 @@ function DiskArrays.writeblock!(a::AbstractPermutedDiskArray, v, i::OrdinalRange
     return nothing
 end
 
-_getperm(::PermutedDiskArray{<:Any,<:Any,perm}) where {perm} = perm
-_getiperm(::PermutedDiskArray{<:Any,<:Any,<:Any,iperm}) where {iperm} = iperm
+_getperm(::AbstractPermutedDiskArray{<:Any,<:Any,perm}) where {perm} = perm
+_getiperm(::AbstractPermutedDiskArray{<:Any,<:Any,<:Any,iperm}) where {iperm} = iperm
 
 # Implementation macro
 


rafaqz:
Maybe ReshapedDiskArray as well for completeness? Not sure if Padded/Rechunked etc also need abstrac types

src/subarray.jl
```diff
@@ -1,32 +1,39 @@
...
+
 """
     SubDiskArray <: AbstractDiskArray
 
 A replacement for `Base.SubArray` for disk arrays, returned by `view`.
 """
-struct SubDiskArray{T,N,P,I,L} <: AbstractDiskArray{T,N}
+struct SubDiskArray{T,N,P,I,L} <: AbstractSubDiskArray{T,N,P,I,L}
     v::SubArray{T,N,P,I,L}
 end
```
lupemba, line 16: I would add a method called 
`get_sub_array(a::AbstractSubDiskArray) = a.v`
and then replace all `a.v` with `get_sub_array(a)`. Using the . notation is not good to extend the type.

src/subarray.jl
```diff
@@ -1,32 +1,39 @@
...
 
 # Base methods
-Base.view(a::SubDiskArray, i...) = SubDiskArray(view(a.v, i...))
-Base.view(a::SubDiskArray, i::CartesianIndices) = view(a, i.indices...)
-Base.size(a::SubDiskArray) = size(a.v)
-Base.parent(a::SubDiskArray) = a.v.parent
+Base.view(a::AbstractSubDiskArray, i...) = SubDiskArray(view(a.v, i...))
```
lupemba, line 18: Here is a idea for a more general method that works for other subtypes
```
function Base.view(a::T, i...) where {T}
    # T.name.wrapper is also used by other packages
    # https://discourse.julialang.org/t/how-to-get-the-container-type-of-a-container/20253/10
    subarray_type = T.name.wrapper
    sub_array = get_sub_array(a) 
    return subarray_type(view(sub_array, i...))
end
```

src/subarray.jl
```diff
@@ -1,32 +1,39 @@
...
+Base.parent(a::AbstractSubDiskArray) = a.v.parent
 
 _replace_colon(s, ::Colon) = Base.OneTo(s)
 _replace_colon(s, r) = r
 
 # Diskarrays.jl interface
-function readblock!(a::SubDiskArray, aout, i::OrdinalRange...)
+function readblock!(a::AbstractSubDiskArray, aout, i::OrdinalRange...)
     pinds = parentindices(view(a.v, i...))
     getindex_disk!(aout, parent(a.v), pinds...)
```
lupemba, line 29: I think you can replace `parent(a.v)` with `parent(a)`

lupemba (commented):
I tried using it to implement to solve https://github.com/JuliaGeo/CommonDataModel.jl/issues/32#issuecomment-2820128429 and here are my comments

src/subarray.jl
tiemvanderdeure: Yeah, I thought about this while writing. `get_sub_array` sounds a little clunky - maybe `parent` should have returned the subarray, but this would be breaking.

rafaqz: I think `parent` need to be the parent data so it's the same as SubArray. But the function can just be `subarray(a)` ?

src/subarray.jl
rafaqz: Probably `where {T<:AbstractSibDiskArray}`

src/subarray.jl
rafaqz: Yes there should be no direct field access now we have abstract types

src/subarray.jl
tiemvanderdeure: on it!

src/subarray.jl
tiemvanderdeure: This just required adding `parentindices` for AbstractSubDiskArray. But fixed now.

src/permute.jl
```diff
@@ -15,8 +15,9 @@ struct PermutedDiskArray{T,N,P<:PermutedDimsArray{T,N}} <: AbstractPermutedDiskA
 end
 
 # Base methods
-Base.size(a::AbstractPermutedDiskArray) = size(a.a)
-Base.parent(a::AbstractPermutedDiskArray) = a.a.parent
+permuteddimsarray(a) = a.a
+Base.size(a::AbstractPermutedDiskArray) = size(permuteddimsarray(a))
+Base.parent(a::AbstractPermutedDiskArray) = parent(permuteddimsarray(a))
```
rafaqz, line 20: We should merge my other PR first, it actually changes this so there is no separate array. Maybe thats a problem?

rafaqz: https://github.com/JuliaIO/DiskArrays.jl/pull/249

tiemvanderdeure: Don't think that's a problem at all! The permuteddimsarray was only necessary to get the perm and iperm anyway.

rafaqz:
@tiemvanderdeure I fixed the merge with #249, it definately makes the abstract type less useful

@lupemba I'm not sure if this will affect you, or if you are subtyping `AbstractPermutedDiskArray`

lupemba:
> @lupemba I'm not sure if this will affect you, or if you are subtyping AbstractPermutedDiskArray

I have left out permuted arrays from my PR since CommonDataModel doesn't have specific methods for that yet. So it is not needed right now.

I still think we should consider adding a support for permuted arrays in CommonDataModel before we make a breaking release.

tiemvanderdeure:
I think it's basically good to go. Let's see if tests pass now.

src/permute.jl
```diff
@@ -41,15 +48,15 @@ function eachchunk(a::PermutedDiskArray)
     # Return permuted GridChunks
     return GridChunks(genperm(gridchunks.chunks, perm)...)
 end
-function DiskArrays.readblock!(a::PermutedDiskArray, aout, i::OrdinalRange...)
+function DiskArrays.readblock!(a::AbstractPermutedDiskArray, aout, i::OrdinalRange...)
     iperm = _getiperm(a)
```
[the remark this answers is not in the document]
tiemvanderdeure: Remove which underscore? You mean in `_getiperm` and `_getperm`

src/permute.jl
```diff
@@ -1,9 +1,16 @@
 """
-    PermutedDiskArray <: AbstractDiskArray
+    AbstractPermutedDiskArray <: AbstractDiskArray
+
+Abstract supertype for diskarray with permuted dimensions.
+"""
+abstract type AbstractPermutedDiskArray{T,N} <: AbstractDiskArray{T,N} end
```
tiemvanderdeure, line 6: Should we actually put all the type parameters of `PermutedDiskArray` in the abstract type as well? We do that with `AbstractReshapedDiskArray`. Then we can define `getperm` on the abstract type.

rafaqz: Sounds good

rafaqz (approved):

PR #254, opened 2025-04-08
Allow missing elements when constructing ConcatDiskArrays

meggart:
This PR contains a more efficient construction of `ConcatDiskArray`s and fixes #248 . In addition, it is now allowed that the array of arrays contains `missing` entries, which means that these represent tiles that contains only missings. These missing values are never allocated but will be inserted on the fly inside the `readblock!` machinery. This might make constructing mosaics of tiles less cumbersome and no need for tricks like rechunking FillArrays or similar (like in https://github.com/meggart/DiskArrayTools.jl/pull/37) just to represent missing values.

Issue #248, opened 2025-03-28
Allow missing tiles in ConcatDiskArray


felixcremer:
When we mosaic DiskArrays together on a >2D grid they're chunks have to align so that the resulting ConcatDiskArray can be constructed. 
Sometimes there are missing tiles and these have to be filled with some value or missing.
For this we introduced ChunkedFillArray in DiskArrayTools so that the chunks can be aligned. But it would be better to allow a missing value or a fill value in the Array of DiskArrays that is the input to ConcatDiskArray.

ConcatDiskArray([1, DiskArray(rand(10,10); DiskArray(rand(10,10), 2] ) should work and would fill the upper box with 1 and the lower box with 2.

rafaqz:
Makes sense, just needs a fill keyword like PaddedDiskArray and handling for readblock where there are missing arrays. I guess that's also kind of similar to PaddedDiskArray

PR #256, opened 2025-04-11
Add functionality to use arbitrary singular values in ConcatDiskArrays

felixcremer:
This adds the possibility to use any value instead of only missing in ConcatDiskArrays. 
We can even mix the singular values.

### DIFF
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -91,7 +91,7 @@ function arraysize_and_startinds(arrays1)
     sizes = map(i -> zeros(Int, i), size(arrays1))
     for i in CartesianIndices(arrays1)
         ai = arrays1[i]
-        !isa(ai, AbstractArray) && continue
+        ai isa AbstractArray || continue
         sizecur = extenddims(size(ai), size(arrays1), 1)
         foreach(sizecur, i.I, sizes) do si, ind, sizeall
             if sizeall[ind] == 0
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -14,7 +14,7 @@ Returned from `cat` on disk arrays.
 
 It is also useful on its own as it can easily concatenate an array of disk arrays.
 """
-struct ConcatDiskArray{T,N,P,C,HC, ID} <: AbstractDiskArray{T,N}
+struct ConcatDiskArray{T,N,P,C,HC,ID} <: AbstractDiskArray{T,N}
     parents::P
     startinds::NTuple{N,Vector{Int}}
     size::NTuple{N,Int}
@@ -33,14 +33,13 @@ end
 
 function infer_eltypes(arrays)
     foldl(arrays, init=(-1, Union{})) do (M, T), a
-        if !isa(a, AbstractArray)
-            (M, promote_type(typeof(a), T))
-        else
+        if a isa AbstractArray
             M == -1 || ndims(a) == M || throw(ArgumentError("All arrays to concatenate must have equal ndims"))
-            (ndims(a), promote_type(eltype(a), T))
+            M = ndims(a)
+        end
+            (M, promote_type(eltype(a), T))
         end
     end
-end
 
 function ConcatDiskArray(arrays::AbstractArray{<:AbstractArray})
     N = ndims(arrays)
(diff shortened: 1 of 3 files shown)


felixcremer:
@rafaqz  this is the functionality that you wanted in #248 . This

src/cat.jl
```diff
@@ -189,13 +191,13 @@ function concat_chunksize(parents)
     newchunks = map(s -> Vector{Union{RegularChunks,IrregularChunks}}(undef, s), size(parents))
     for i in CartesianIndices(parents)
         array = parents[i]
-        ismissing(array) && continue
+        !isa(array,AbstractArray) && continue
```
rafaqz, line 194:
```suggestion
        array isa AbstractArray || continue
```

Maybe easier to read

src/cat.jl
```diff
@@ -90,7 +91,7 @@ function arraysize_and_startinds(arrays1)
     sizes = map(i -> zeros(Int, i), size(arrays1))
     for i in CartesianIndices(arrays1)
         ai = arrays1[i]
-        ismissing(ai) && continue
+        !isa(ai, AbstractArray) && continue
```
rafaqz, line 94:
```suggestion
        ai isa AbstractArray || continue
```

src/cat.jl
```diff
@@ -123,10 +124,11 @@ function readblock!(a::ConcatDiskArray, aout, inds::AbstractUnitRange...)
     # Find affected blocks and indices in blocks
     _concat_diskarray_block_io(a, inds...) do outer_range, array_range, I
         vout = view(aout, outer_range...)
-        if ismissing(I)
-            vout .= missing
-        else
+        #@show size(vout)
```
rafaqz, line 127:
```suggestion
```

src/cat.jl
```diff
@@ -24,23 +23,25 @@ struct ConcatDiskArray{T,N,P,C,HC,ID} <: AbstractDiskArray{T,N}
     innerdims::Val{ID}
 end
 
-function ConcatDiskArray(arrays::AbstractArray{Union{<:AbstractArray,Missing}})
+function ConcatDiskArray(arrays::AbstractArray{Union{<:AbstractArray,Missing}}; fill=missing)
     et = Base.nonmissingtype(eltype(arrays))
-    T = Union{Missing,eltype(et)}
+    T = promotetype(typeof(fill), eltype(et))
```
rafaqz, line 28: I thought the passed in array would still contain either arrays or `missing` values, but we would just return `fill` for those sections, so `fill` would need to be a struct field.

So what we have here is kinda half way. Probably we should either drop the `fill` keyword completely or keep it as a struct field and use `missing` in the array.

src/cat.jl
felixcremer: True

src/cat.jl
felixcremer: I prefer to drop the fill keyword. I actual thought I already did so. 
With my approach we can have different fill values in different parts of the concatenated array.

src/cat.jl
```diff
@@ -24,23 +23,25 @@ struct ConcatDiskArray{T,N,P,C,HC,ID} <: AbstractDiskArray{T,N}
     innerdims::Val{ID}
 end
 
-function ConcatDiskArray(arrays::AbstractArray{Union{<:AbstractArray,Missing}})
+function ConcatDiskArray(arrays::AbstractArray{Union{<:AbstractArray,Missing}}; fill=missing)
```
felixcremer, line 26:
```suggestion
function ConcatDiskArray(arrays::AbstractArray{Union{<:AbstractArray,Missing}})
```

src/cat.jl
felixcremer, line 28:
```suggestion
    T = Union{Missing,eltype(et)}
```

felixcremer:
Why do the tests not run on this pr?

src/cat.jl
rafaqz: But is that actually useful? I can't imagine a real use case for multiple fill values. And putting `missing` in the array seems clearer for representing missing arrays. 

The missing value used in the array of arrays doesn't have to be coupled to the fill value.

meggart: I agree that it might be a bit dangerous to allow arbitrary values to represent missings. OTOH one never knows when this might become useful for some reason. I am ok with both ways and would currently slightly prefer @felixcremer s solution because there is already an implementation. If we want to be a bit more on the safe side, we could introduce a small wrapper type like

````julia
struct MissingTile{F}
    fillvalue::F
end
````
that explicitly marks missing entries to be concatenated and how they should be filled. The main reason why the current solution might be problematic is for DiskArrays with arrays as eltype, but maybe this is overthinking things a bit too much.

felixcremer:
I took this for a spin and I am not sure, whether it is this change or #256 or ConcatDiskArray even before but saving a ConcatDiskArray where I have some missings takes a very long time. Saving the european dataset took the whole weekend.

src/cat.jl
rafaqz: Don't you have to read all the values and promote them? Missing is not necessarily in the type?

meggart:
I will merge this for now, as it is only a PR against my other PR. We can continue discussing in #254

PR #260, opened 2025-04-21
fix and test mixed CartesianIndex

### LINKED ISSUE #259: Mixed indexing fail. a[CartesianIndex(1, 2), 1:end]
Missing CartesianIndex with a range gives an error.

```julia
using DiskArrays: ReshapedDiskArray, PermutedDiskArray
using DiskArrays.TestTypes
data = reshape(1:24, 2, 3, 4)
a = AccessCountDiskArray(data, chunksize=(2, 2, 2))

a[CartesianIndex(1, 2), 1:end] # gives an error
data[CartesianIndex(1, 2), 1:end] # works fine
```

rafaqz:
Fixes #259

@lupemba hopefully this solves your issue

### DIFF
--- a/src/diskindex.jl
+++ b/src/diskindex.jl
@@ -182,7 +182,12 @@ splitchunks(i::CartesianIndex, chunks) = splitchunks(i.I, (), chunks)
 splitchunks(_, chunks) = (first(chunks),), Base.tail(chunks)
 splitchunks(si, chunksnow, chunksrem) =
     splitchunks(Base.tail(si), (chunksnow..., first(chunksrem)), Base.tail(chunksrem))
+function splitchunks(si,chunksnow, ::Tuple{})
+    only(first(si)) == 1 || throw(ArgumentError("Trailing indices must be 1"))
+    splitchunks(Base.tail(si), chunksnow, ())
+end
 splitchunks(::Tuple{}, chunksnow, chunksrem) = (chunksnow, chunksrem)
+splitchunks(::Tuple{}, chunksnow, chunksrem::Tuple{}) = (chunksnow, chunksrem)
 
 """
     output_aliasing(di::DiskIndex, ndims_dest, ndims_source)
--- a/src/diskindex.jl
+++ b/src/diskindex.jl
@@ -52,7 +52,13 @@ DiskIndex(a, i::Tuple{<:AbstractVector{<:Integer}}, batchstrategy) =
 function _resolve_indices(chunks, i, indices_pre::DiskIndex, strategy::BatchStrategy)
     inow = first(i)
     indices_new, chunksrem = process_index(inow, chunks, strategy)
-    _resolve_indices(chunksrem, Base.tail(i), merge_index(indices_pre, indices_new), strategy)
+    _resolve_indices(chunksrem, tail(i), merge_index(indices_pre, indices_new), strategy)
+end
+# Splat out CartesianIndex as regular indices
+function _resolve_indices(
+    chunks, i::Tuple{<:CartesianIndex}, indices_pre::DiskIndex, strategy::BatchStrategy
+)
+    _resolve_indices(chunks, (Tuple(i[1])..., tail(i)...), indices_pre, strategy)
 end
 _resolve_indices(::Tuple{}, ::Tuple{}, indices::DiskIndex, strategy::BatchStrategy) = indices
 # No dimension left in array, only singular indices allowed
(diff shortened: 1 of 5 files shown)


lupemba:
I just did a small test and discovered that it doesn't work for the views. 
```julia
using DiskArrays: ReshapedDiskArray, PermutedDiskArray
using DiskArrays.TestTypes

a = AccessCountDiskArray(reshape(1:24, 2, 3, 4), chunksize=(2, 2, 2))
a[CartesianIndex(1, 2), 1:end] # works now
@view a[CartesianIndex(1, 2), 1:end] # fails
```
I don't think I need the functionality but I just wanted to share it right away before I forget.

rafaqz:
An the latest push? I tried to fix view too, it wirks in the tests now. Have to check why this argument combo doesn't work.

lupemba:
Yes in the latest commit on this branch. The error comes from "DiskArrays.jl/src/subarray.jl:65"

`i2 = _replace_colon.(size(a), i)` size(a) has length 3 and i only has length 2 so broadcasting fails.

```
ERROR: DimensionMismatch: arrays could not be broadcast to a common size: a has axes Base.OneTo(3) and b has axes Base.OneTo(2)
Stacktrace:
  [1] _bcs1
    @ ./broadcast.jl:528 [inlined]
  [2] _bcs
    @ ./broadcast.jl:522 [inlined]
  [3] broadcast_shape(::Tuple{Base.OneTo{Int64}}, ::Tuple{Base.OneTo{Int64}})
    @ Base.Broadcast ./broadcast.jl:516
  [4] combine_axes
    @ ./broadcast.jl:497 [inlined]
  [5] _axes
    @ ./broadcast.jl:236 [inlined]
  [6] axes
    @ ./broadcast.jl:234 [inlined]
  [7] copy
    @ ./broadcast.jl:1099 [inlined]
  [8] materialize
    @ ./broadcast.jl:872 [inlined]
  [9] view(::AccessCountDiskArray{…}, ::CartesianIndex{…}, ::UnitRange{…})
    @ DiskArrays ~/Documents/git repos/DiskArrays.jl/src/subarray.jl:65
 [10] top-level scope
    @ ~/Documents/git repos/DiskArrays.jl/small_test.jl:6
Some type information was truncated. Use `show(err)` to see complete types.
```

rafaqz:
Ok thanks, I'll fix that tomorrow or monday

rafaqz:
Ok so it turns out this fails with any array at all! Its just julias `end` doesn't work with CartesianIndex in the mix, it will lower to `lastindex(a, 2)` rather than `lastindex(a, 3)`.

```julia
@view a[CartesianIndex(1, 2), 1:end] # fails
```

rafaqz:
@meggart could you have a look at this? it would be good to merge for @lupemba s dependent PR

lupemba:
> Ok so it turns out this fails with any array at all! Its just julias end doesn't work with CartesianIndex in the mix, it will lower to lastindex(a, 2) rather than lastindex(a, 3).

Interesting, I will add an issue to the Julia repository to document this.

```julia
a = reshape(1:(2*3*4),(4,3,2))

a[CartesianIndex(1, 2), :] # works
a[1,2, 1:end] # works
a[CartesianIndex(1, 2), 1:end] # fails 
```

src/indexing.jl
```diff
@@ -56,6 +56,10 @@ function getindex_disk(a::AbstractArray, i::Integer)
     return only(outputarray)
 end
 getindex_disk(a::AbstractArray, i...) = getindex_disk!(nothing, a, i...)
+getindex_disk(a::AbstractArray, i::ChunkIndex{<:Any,OneBasedChunks}) =
+    a[eachchunk(a)[i.I]...]
+getindex_disk(a::AbstractArray, i::ChunkIndex{<:Any,OffsetChunks}) =
```
meggart, line 61: Not sure why these new methods are necessary? Or have they simply been moved?

src/indexing.jl
```diff
@@ -306,10 +310,6 @@ macro implement_getindex(t)
     quote
         DiskArrays.isdisk(::Type{<:$t}) = true
         Base.getindex(a::$t, i...) = getindex_disk(a, i...)
-        @inline Base.getindex(a::$t, i::ChunkIndex{<:Any,OneBasedChunks}) =
-            a[eachchunk(a)[i.I]...]
-        @inline Base.getindex(a::$t, i::ChunkIndex{<:Any,OffsetChunks}) =
-            wrapchunk(a[nooffset(i)], eachchunk(a)[i.I])
```
meggart, line 312: Ah I see, it is reorganisation

src/diskindex.jl
```diff
@@ -159,10 +178,15 @@ splitchunks(i::AbstractArray{<:CartesianIndex}, chunks) =
...
+splitchunks(_, chunks) = (first(chunks),), tail(chunks)
 splitchunks(si, chunksnow, chunksrem) =
-    splitchunks(Base.tail(si), (chunksnow..., first(chunksrem)), Base.tail(chunksrem))
+    splitchunks(tail(si), (chunksnow..., first(chunksrem)), tail(chunksrem))
+function splitchunks(si, chunksnow, chunksrem::Tuple{})
+    checktrailing(si[1])
+    splitchunks(tail(si), chunksnow, chunksrem)
+end
 splitchunks(::Tuple{}, chunksnow, chunksrem) = (chunksnow, chunksrem)
+splitchunks(::Tuple{}, chunksnow, chunksrem::Tuple{}) = (chunksnow, chunksrem)
```
meggart, line 189: I am not sure why modifications here are necessary. Could you maybe try again without this new methods. I have a slight suspicion this is only necessary because the implementation of `process_index` above is not correct.

meggart (changes requested):
Thanks a lot for digging into this. Looks good in general, except the implementation of `process_index`

src/indexing.jl
rafaqz: I just moved them out of the macro for Revise.jl, along with view methods etc

(Had that situation where you need to reload Julia to get code updates)

src/diskindex.jl
```diff
@@ -98,18 +112,23 @@ Calculate indices for `i` the first chunk/s in `chunks`
 Returns a [`DiskIndex`](@ref), and the remaining chunks.
 """
 process_index(i, chunks, ::NoBatch) = process_index(i, chunks)
-process_index(inow::Integer, chunks) = DiskIndex((), (1,), (), (1,), (inow:inow,)), Base.tail(chunks)
+function process_index(I::CartesianIndex, chunks, bs::NoBatch)
+    _, chunksrem = splitchunks(I, chunks)
+    DiskIndex((), (1, 1), (), (1, 1), map(i -> i:i, Tuple(I))), chunksrem
```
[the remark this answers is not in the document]
rafaqz: fixed

src/diskindex.jl
meggart: Thanks for reverting, I will investigate from here

meggart:
Ok, I finally understood the bug. It was not view-related but could be induced by a getindex on a normal DiskArray as well when indexing with a longer `CartesianIndex` than there were actual dimensions. I re-implemented the fix with a (probably unnecessary) additional argument check. Thanks for taking care.

PR #264, opened 2025-06-22
Add 5 arg `copyto!` into `Vector` with matching type

### LINKED ISSUE #263: Scalar indexing in `function copyto!(dest::Array, doffs::Integer, src::AbstractDiskArray, soffs::Integer, n::Integer)`
The current 5-argument `copyto!` uses scalar indexing.

  ```julia-repl
julia> using DiskArrays

julia> DiskArrays.allowscalar(false);

julia> a = ones(UInt8, 24);

julia> da = AccessCountDiskArray(a; chunksize=(4,));

julia> copyto!(zeros(UInt8,3), 1, da, 5, 3)
ERROR: Scalar indexing with `Int` is very slow, and currently is disallowed. Run DiskArrays.allowscalar(true) to allow
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] _scalar_error()
   @ DiskArrays ~/github/DiskArrays.jl/src/scalar.jl:34
 [3] checkscalar
   @ ~/github/DiskArrays.jl/src/scalar.jl:30 [inlined]
 [4] checkscalar
   @ ~/github/DiskArrays.jl/src/scalar.jl:31 [inlined]
 [5] getindex_disk
   @ ~/github/DiskArrays.jl/src/indexing.jl:47 [inlined]
 [6] getindex
   @ ~/github/DiskArrays.jl/src/indexing.jl:312 [inlined]
 [7] copyto!(dest::Vector{UInt8}, dstart::Int64, src::AccessCountDiskArray{UInt8, 1, Vector{UInt8}, DiskArrays.ChunkRead{DiskArrays.NoStepRange}}, sstart::Int64, n::Int64)
   @ Base ./abstractarray.jl:1128
 [8] top-level scope
   @ REPL[31]:1

```

Fixing this will greatly improve performance using the [reading functions](https://github.com/JuliaIO/InputBuffers.jl/blob/b2718478ad032a78737cd249cc07f094c29a4123/src/InputBuffers.jl#L81-L90) on an `AbstractDiskArray` wrapped in an `InputBuffer` from: https://github.com/JuliaIO/InputBuffers.jl

Here is an example of what FixedSizeArray does to avoid ambiguity:
https://github.com/JuliaArrays/FixedSizeArrays.jl/blob/35fb27aac19e0cc66ae001c49202c538c4033a56/src/FixedSizeArray.jl#L359-L382

### PR DESCRIPTION
**Pull request by nhz2**
Fixes #263

I only implemented the one-dimensional case because it wasn't obvious to me how to handle multiple dimensions, and this is all that is required for better `InputBuffers.jl` compatibility for now.

I also modified the `copyto!` function to always return `dest`, which somehow fixed the `circshift` test.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -23,7 +23,7 @@ function _disk_copyto!(dest, Rdest, src, Rsrc)
     view(dest, Rdest) .= view(src, Rsrc)
     return dest
 end
-function _disk_copyto_vector!(dest, dstart, src, sstart, n)
+function _disk_copyto_same_type_vector!(dest, dstart, src, sstart, n)
     if iszero(n)
         return dest
     end
@@ -33,6 +33,17 @@ function _disk_copyto_vector!(dest, dstart, src, sstart, n)
     end
     destv = view(dest, range(dstart, length=n))
     DiskArrays.readblock!(src, destv, range(sstart, length=n))
+    return dest
+end
+function _disk_copyto_vector!(dest, dstart, src, sstart, n)
+    if iszero(n)
+        return dest
+    end
+    if n < 0
+        throw(ArgumentError(LazyString("tried to copy n=",
+        n," elements, but n should be non-negative")))
+    end
+    view(dest, range(dstart, length=n)) .= view(src, range(dstart, length=n))
     return dest
 end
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/array.jl [hunk 1]
```diff
@@ -76,6 +89,9 @@ macro implement_array_methods(t)
         function Base.copyto!(dest::PermutedDimsArray{T,N}, src::$t{T,N}) where {T,N}
             return $_disk_copyto!(dest, src)
         end
+        function Base.copyto!(dest::Vector{T}, dstart::Integer, src::$t{T, 1}, sstart::Integer, n::Integer) where {T}
```
rafaqz, line 92: Is `T` always identical for scr and dest in Bae `copyto!` ? What if you have different Real types

rafaqz: We also use `setindex!` when wtiting into the chunk buffer, so it should be the same?

nhz2:
I removed the specialization that does type conversion to limit the scope of this PR.

src/array.jl
```diff
@@ -21,6 +21,30 @@ function _disk_copyto!(dest, Rdest, src, Rsrc)
...
+end
+function _disk_copyto_vector!(dest, dstart, src, sstart, n)
+    if iszero(n)
+        return dest
+    end
+    if n < 0
+        throw(ArgumentError(LazyString("tried to copy n=",
+        n," elements, but n should be non-negative")))
+    end
+    view(dest, range(dstart, length=n)) .= view(src, range(dstart, length=n))
```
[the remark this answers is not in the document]
rafaqz: I would also prefer @meggarts approach, although you can use a view rather than get index on `src`. But let's try to avoid nested disk broadcasts

rafaqz:
I would like this to work on e.g. Float64 dest from a Float 32 diskarray.

Dest is an `Array` so it calls `convert`. Just a few Ts to delete.

PR #275, opened 2025-08-31
only call `sizeof` on isbits types

### LINKED ISSUE #261: Dont call `sizeof` on Strings and other types with unknown size
https://github.com/JuliaIO/DiskArrays.jl/blob/80274cd3679f90c04c4338fa7c5dfbd0089284d4/src/broadcast.jl#L65

This will fail on `String`

rafaqz:
This currently fails for `String`, and will for any other non-isbits type

This PR adds another path for non-isbits types in broadcast that just treats the whole array as one chunk.

Closes #261

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -64,8 +64,7 @@ function common_chunks(s, args...)
     if isempty(chunkedarrays)
         # Estimate chunk size for isbits
         if all(map(isbits ∘ eltype, args))
-            totalsize = sum(sizeof ∘ eltype, args)
-            return estimate_chunksize(s, totalsize)
+            return estimate_chunksize(s)
         else # Otherwise just use one huge chunk, we dont know what the object is
             return GridChunks(s, s)
         end
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -62,15 +62,20 @@ function common_chunks(s, args...)
     all(ar -> isa(eachchunk(ar), GridChunks), chunkedarrays) ||
         error("Currently only chunks of type GridChunks can be merged by broadcast")
     if isempty(chunkedarrays)
-        totalsize = sum(sizeof ∘ eltype, args)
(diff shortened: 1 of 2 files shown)


src/broadcast.jl
```diff
@@ -62,15 +62,20 @@ function common_chunks(s, args...)
     all(ar -> isa(eachchunk(ar), GridChunks), chunkedarrays) ||
         error("Currently only chunks of type GridChunks can be merged by broadcast")
     if isempty(chunkedarrays)
-        totalsize = sum(sizeof ∘ eltype, args)
-        return estimate_chunksize(s, totalsize)
```
meggart, line 66: I think this function does actually the wrong thing. Can we just call into this method https://github.com/JuliaIO/DiskArrays.jl/blob/877fd67b3374089ed30a4c937c2db97b4af22dd6/src/chunks.jl#L441 and have the non-isbits types covered?

rafaqz: @meggart I honestly don't know what you mean here. Can you just edit it into the code as a suggestion thats usually clearer.

This is breaking things in Rasters.jl and this comment is blocking from merging a fix that works

asinghvi17: Can't suggest for a deleted line

rafaqz: No you but you can suggest on the other side?

src/broadcast.jl
```diff
@@ -62,15 +62,20 @@ function common_chunks(s, args...)
...
+        # Estimate chunk size for isbits
+        if all(map(isbits ∘ eltype, args))
+            totalsize = sum(sizeof ∘ eltype, args)
+            return estimate_chunksize(s, totalsize)
```
asinghvi17:
```suggestion
            return estimate_chunksize(s)
```

asinghvi17: this is (I think) what @meggart was talking about

PR #276, opened 2025-09-15
Allow mockchunks to accept any AbstractArray

asinghvi17:


meggart (approved):

asinghvi17:
https://github.com/rafaqz/DimensionalData.jl/actions/runs/18267265353/job/52003484300#step:5:1412

tiemvanderdeure:
Doesn't this just totally break `mockchunks`? It now calls `GridChunks(size(data), chunks)`, but we don't have a dispatch of `GridChunks(::Tuple, ::GridChunks)`

PR #278, opened 2025-10-07
Allow `GridChunks` in mockchunks but without potential ambiguity

asinghvi17:
Fix the breakage from #276 

@rafaqz @tiemvanderdeure

### DIFF
--- a/src/mockchunks.jl
+++ b/src/mockchunks.jl
@@ -33,7 +33,7 @@ function mockchunks(data::AbstractArray, chunks)
     else
         GridChunks(data, chunks)
     end
-    MockChunkedDiskArray(data, GridChunks(data, chunks))
+    MockChunkedDiskArray(data, gridchunks)
 end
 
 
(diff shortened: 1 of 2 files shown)


src/mockchunks.jl
```diff
@@ -27,7 +27,15 @@ The chunks argument can take one of the following forms:
 - a [`DiskArrays.GridChunks`](@ref) object
 - a tuple specifying the chunk size along each dimension, like `(10, 10, 1)` for a 3-D array
 """
-mockchunks(data::AbstractArray, chunks) = MockChunkedDiskArray(data, GridChunks(size(data), chunks))
+function mockchunks(data::AbstractArray, chunks)
```
felixcremer, line 30: Shouldn't we be doing that via Dispatch on chunks?

asinghvi17: I want to leave room to dispatch on the first argument, and not have ambiguities if the second argument is typed as well.  DD does override `mockchunks(::AbstractDimArray, chunks)`, which would introduce an ambiguity if you ever did `mockchunks(::AbstractDimArray, ::GridChunks)` if we did this by dispatch.

It should be equally fast but maybe I can add a specialization to make sure.

```suggestion
function mockchunks(data::AbstractArray, chunks::C) where C
```

felixcremer:
I just added a fix and tests to this as a PR in #279

asinghvi17:
Looks like this is ready to merge

PR #280, opened 2025-10-20
Remove singleton dimensions in reshape and add dropdims

### LINKED ISSUE #184: `dropdims` should be able to remove trailing 1s
Currently calling `dropdims(A, n)` on a disk array gives the error: 

`"For DiskArrays, reshape is restricted to adding singleton dimensions"`

We should define `dropdims` to just call `view` instead of `reshape` - this is what Rasters.jl does manually - but it would be nice if `dropdims` just worked.

meggart:
Fixes #184 @lazarusA @felixcremer

### DIFF
--- a/src/reshape.jl
+++ b/src/reshape.jl
@@ -92,13 +92,6 @@ function reshape_index(a, default, replace)
     inew
 end
 
-#This is currently not type-stable, feel free to make an improved version through a PR
-function dropdims_disk(a; dims)
-    remaining_dims = setdiff(1:ndims(a), dims)
-    newshape = size(a)[remaining_dims]
-    reshape(a, newshape)
-end
-
 
 # Implementaion macro
 macro implement_reshape(t)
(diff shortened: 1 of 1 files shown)


src/reshape.jl
```diff
@@ -20,73 +20,85 @@ rectangles in the parent array.
...
+        if m != -1
+            indx = replace[ii]
+            inew = Base.setindex(inew, indx, m)
+        end
+    end
+    inew
+end
+
+#This is currently not type-stable, feel free to make an improved version through a PR
+function dropdims_disk(a; dims)
```
felixcremer: Is this even necessary? 
I tested removing the Base.dropdims overload below and it still works as expected and the dropdims tests still pass. This works ,because the Base.dropdims fallback definition does some size checks and then calls rebase on the dimensions.

meggart: Great! I removed the dropdims implementation.

felixcremer (approved):
LGTM

PR #283, opened 2025-10-24
Add TiledDiskArray

felixcremer:
Add a TiledDiskArray type which handles a collection of tiles that lay in the same grid with a predictable pattern .

### DIFF
--- a/src/chunktiledarray.jl
+++ b/src/chunktiledarray.jl
@@ -8,12 +8,14 @@ abstract type AbstractChunkTiledDiskArray{T,N} <: AbstractDiskArray{T,N} end
 
 Base.size(a::AbstractChunkTiledDiskArray) = arraysize_from_chunksize.(eachchunk(a).chunks)
 
-function readblock!(A::AbstractAbstractChunkTiledDiskArray{T,N}, data, I...) where {T,N}
+function readblock!(A::AbstractChunkTiledDiskArray{T,N}, data, I...) where {T,N}
     chunks = eachchunk(A)
     chunk_indices = findchunk.(chunks.chunks, I)
     data_offset = OffsetArray(data, map(i -> first(i) - 1, I)...)
     foreach(CartesianIndices(chunk_indices)) do ci
+        @show ci
         chunkindex = ChunkIndex(ci; offset=true)
+        @show chunkindex
         chunk = A[chunkindex]
         # Find the overlapping indices
         inner_indices = map(axes(chunk), axes(data_offset)) do ax1, ax2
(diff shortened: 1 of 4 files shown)


rafaqz:
Could we use this to lazily wrap a tile server?

felixcremer:
Main problem might be, that a Tile Server has color values and not actual array values. But this might work for wrapping a tile server and giving the RGB values from the pngs. 

This would currently expect a function that would from a given tile position return a matrix of the data in this position. So we could have a function that reads the data from the tile server. 

The interface we are aiming for currently is this:

```julia
using ArchGDAL

function geturl(east, north) 
    baseurl = "https://download.geoservice.dlr.de/FNF50/files"
    xsym, xval = if east < 180
        'W',180-east
    else
        'E', east-180
    end
    ysym, yval = if north < 90
        'S', 90-north
    else
        'N', north-90
    end
    xpadround = xsym * lpad(Int(round(xval, RoundDown, digits=-1)), 3, "0")
    ypadround = ysym * lpad(Int(round(yval, RoundDown, digits=-1)), 2, "0")
    xpad = xsym * lpad(xval, 3, "0")
    ypad = ysym * lpad(yval, 2,"0")
    tdm = "TDM_FNF_20_" * ypad * xpad
    join([baseurl, ypad, xpadround, tdm, tdm * "_cog.tif"], "/")
end

function gettile(east, north)
    response = HTTP.head(geturl(1,1), status_exception=false)
    if response.status == 200
        ArchGDAL.readraster("/vsicurl/"*geturl(east, north))[:,:]::Matrix{UInt8}
    else
        fill(UInt8(3), (2001,2001))
    end
end

A = DiskArrays.TiledDiskArray{UInt8,2}(gettile, (360,180), (2001,2001))
```

src/chunktiledarray.jl
```diff
@@ -0,0 +1,60 @@
...
+abstract type AbstractChunkTiledDiskArray{T,N} <: AbstractDiskArray{T,N} end
+
+Base.size(a::AbstractChunkTiledDiskArray) = arraysize_from_chunksize.(eachchunk(a).chunks)
+
+function readblock!(A::AbstractChunkTiledDiskArray{T,N}, data, I...) where {T,N}
+    chunks = eachchunk(A)
+    chunk_indices = findchunk.(chunks.chunks, I)
+    data_offset = OffsetArray(data, map(i -> first(i) - 1, I)...)
+    foreach(CartesianIndices(chunk_indices)) do ci
+        @show ci
```
meggart, line 16: remove

src/chunktiledarray.jl
```diff
@@ -0,0 +1,60 @@
...
+        chunkindex = ChunkIndex(ci; offset=true)
+        @show chunkindex
```
meggart, line 18: remove

src/chunktiledarray.jl
```diff
@@ -0,0 +1,60 @@
...
+"""
+    TiledDiskArray <: AbstractChunkTiledDiskArray
+
+Construct an array from a collection of tiles. 
+This needs a function to find the tile given a tile position and the overall size of the array.
+"""
+struct TiledDiskArray{T,N} <: AbstractChunkTiledDiskArray{T,N}
+    tilefunction
+    tilenum
+    tilesizes
```
meggart, line 39: Can you use this definition instead:

````julia
struct TiledDiskArray{T,N,F} <: AbstractChunkTiledDiskArray{T,N}
    tilefunction::F
    tilenum::NTuple{N,Int}
    tilesizes::NTuple{N,Int}
end
````

meggart (changes requested):

src/chunktiledarray.jl
asinghvi17: Might be worth having the tile definition be GridChunks instead - for example the Copernicus DEM has irregular sizes

felixcremer: Do you have a link to that?

asinghvi17: https://copernicus-dem-30m.s3.amazonaws.com/readme.html

rafaqz, line 39: Is `tilesizes` just `tilesize`? What does the pluralisation mean?

meggart:
Superseded by #300 sorry @felixcremer, I could not push to your branch

PR #286, opened 2026-01-19
Implement similar and copy for 0-dimensional arrays

### LINKED ISSUE #285: Subtracting 0-dimensional AbstractDiskArrays fails
## The problem

Here's an MWE:
```julia

julia> v = UnchunkedDiskArray(rand(2));

julia> v-v;  # works

julia> a = UnchunkedDiskArray(fill(rand()));

julia> a - a; # fails
ERROR: MethodError: no method matching similar(::Base.Broadcast.Broadcasted{…}, ::Type{…}, ::Tuple{})
The function `similar` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  similar(::Base.ReshapedArray, ::Type, ::NTuple{N, Int64} where N)
   @ Base reshapedarray.jl:252
  similar(::Array, ::Type, ::NTuple{N, Int64}) where N
   @ Base array.jl:377
  similar(::Base.ReinterpretArray, ::Type, ::NTuple{N, Int64} where N)
   @ Base reinterpretarray.jl:223
  ...

Stacktrace:
 [1] similar(bc::Base.Broadcast.Broadcasted{DiskArrays.ChunkStyle{…}, Nothing, typeof(-), Tuple{…}}, ::Type{DiskArrays.BroadcastDiskArray{…}})
   @ Base.Broadcast ./broadcast.jl:227
 [2] broadcast_preserving_zero_d
   @ ./broadcast.jl:884 [inlined]
 [3] -(A::UnchunkedDiskArray{Float64, 0, Array{Float64, 0}}, B::UnchunkedDiskArray{Float64, 0, Array{Float64, 0}})
   @ Base ./arraymath.jl:8
 [4] top-level scope
   @ REPL[8]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

Same error happens for `ChunkedDiskArray`.

## Proposed solution

This happens because subtracting `AbstractArray`s ultimately hits [this method](https://github.com/JuliaLang/julia/blob/ba1e628ee49351af0b704afd2b2903d253bd3564/base/broadcast.jl#L881-L885), which for 0-dimensional arrays calls `similar(bc, ::Type{T})` on the result of `Base.broadcasted(-, a, a)`. For a normal `Array`, we hit these [two methods](https://github.com/JuliaLang/julia/blob/ba1e628ee49351af0b704afd2b2903d253bd3564/base/broadcast.jl#L228-L231), but DiskArrays lacks these analogous methods:

```julia
Base.similar(::Base.Broadcast.Broadcasted{DiskArrays.ChunkStyle{N}}, ::Type{ElType}, dims) where {N,ElType} =
    similar(Array{ElType, length(dims)}, dims)
Base.similar(::Base.Broadcast.Broadcasted{DiskArrays.ChunkStyle{N}}, ::Type{Bool}, dims) where N =
    similar(BitArray, dims)
```

This isn't quite enough to make the above example work, because DiskArrays also lacks an analog to [this method](https://github.com/JuliaLang/julia/blob/ba1e628ee49351af0b704afd2b2903d253bd3564/base/broadcast.jl#L909):

```julia
@inline Base.copy(bc::Base.Broadcast.Broadcasted{DiskArrays.ChunkStyle{0}}) = bc[CartesianIndex()]
```

With these 3 overloads, the example works:
```julia
julia> v - v
2-element DiskArrays.BroadcastDiskArray{Float64, 1, Base.Broadcast.Broadcasted{DiskArrays.ChunkStyle{1}, Tuple{Base.OneTo{Int64}}, typeof(-), Tuple{UnchunkedDiskArray{Float64, 1, Vector{Float64}}, UnchunkedDiskArray{Float64, 1, Vector{Float64}}}}}

Chunked: (
    [2]
)


julia> a - a
0-dimensional Array{Float64, 0}:
0.0
```

Does it make sense to add these?

sethaxen:
Fixes #285

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -42,6 +42,13 @@ function Base.copyto!(dest::AbstractArray, broadcasted::Broadcasted{ChunkStyle{N
     return dest
 end
 
+function Base.similar(::Broadcasted{ChunkStyle{0}}, ::Type{ElType}, dims) where {ElType}
+    return similar(Array{ElType, length(dims)}, dims)
+end
+function Base.similar(::Broadcasted{ChunkStyle{0}}, ::Type{Bool}, dims)
+    return similar(BitArray, dims)
+end
+
 # DiskArrays interface
 
 haschunks(a::BroadcastDiskArray) = Chunked()
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -28,6 +28,7 @@ Base.copy(a::BroadcastDiskArray) = copyto!(zeros(eltype(a), size(a)), a.broadcas
 
 Base.copy(broadcasted::Broadcasted{ChunkStyle{N}}) where {N} =
     BroadcastDiskArray(flatten(broadcasted))
+@inline Base.copy(broadcasted::Broadcasted{ChunkStyle{0}}) = broadcasted[CartesianIndex()]
 function Base.copyto!(dest::AbstractArray, broadcasted::Broadcasted{ChunkStyle{N}}) where {N}
     bcf = flatten(broadcasted)
     # Get a list of chunks to apply
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -359,15 +359,15 @@ end
 
 # https://github.com/JuliaIO/DiskArrays.jl/issues/285
 @testset "basic array ops work for 0-dimensional arrays" begin
-    aval = 2
-    bval = 3
-    a = UnchunkedDiskArray(fill(aval))
-    b = UnchunkedDiskArray(fill(bval))
-    @testset for op in (-, +)
-        cval = op(aval, bval)
-        c = @inferred op(a, b)
-        @test c isa Array{typeof(cval),0}
-        @test c[] == cval
+    @testset "$(typeof(aval))" for (aval, bval) in ((2, 3), (true, false))
+        a = UnchunkedDiskArray(fill(aval))
+        b = UnchunkedDiskArray(fill(bval))
+        @testset for op in (-, +)
+            cval = op(aval, bval)
+            c = @inferred op(a, b)
+            @test c isa Array{typeof(cval),0}
+            @test c[] == cval
+        end
     end
 end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -357,6 +357,20 @@ end
     @test a[] == 6
 end
 
+# https://github.com/JuliaIO/DiskArrays.jl/issues/285
+@testset "basic array ops work for 0-dimensional arrays" begin
+    aval = 2
+    bval = 3
+    a = UnchunkedDiskArray(fill(aval))
+    b = UnchunkedDiskArray(fill(bval))
+    @testset for op in (-, +)
+        cval = op(aval, bval)
+        c = @inferred op(a, b)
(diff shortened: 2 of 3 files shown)


asinghvi17:
Seems good to me but I'm not super knowledgable about the codebase, would be good for @felixcremer or @meggart to chime in here

test/runtests.jl
```diff
@@ -357,6 +357,28 @@ end
...
 
+# https://github.com/JuliaIO/DiskArrays.jl/issues/285
+@testset "basic array ops work for 0-dimensional arrays" begin
+    @testset "Int" begin
+        aval, bval = 2, 3
+        a = UnchunkedDiskArray(fill(aval))
+        b = UnchunkedDiskArray(fill(bval))
+        @testset for op in (-, +)
+            cval = op(aval, bval)
+            c = @inferred op(a, b)
```
rafaqz, line 368: Shouldnt this be a broadcast?

meggart: Good point, I am wondering the same. Does generic `+` and `-` for AbstractArrays fall back to calling broadcast?

felixcremer: generic + and - fall back to calling broadcast_preserving_zero_d for AbstractArrays

sethaxen: > Shouldnt this be a broadcast?

No, because broadcasting 2 0-dimensional arrays doesn't produce a 0-dimensional array, and it turns out this doesn't hit these functions at all:

```julia
julia> typeof(fill(3.0) + fill(4.0))
Array{Float64, 0}

julia> typeof(fill(3.0) .+ fill(4.0))
Float64
```

> Good point, I am wondering the same. Does generic `+` and `-` for AbstractArrays fall back to calling broadcast?

Yes, see #285 for the specific methods that are hit. Anything that ultimately calls `broadcast_preserving_zero_d` on a 0-dimensional array will hit these methods.

src/broadcast.jl
```diff
@@ -42,6 +43,13 @@ function Base.copyto!(dest::AbstractArray, broadcasted::Broadcasted{ChunkStyle{N
     return dest
 end
 
+function Base.similar(::Broadcasted{ChunkStyle{0}}, ::Type{ElType}, dims) where {ElType}
+    return similar(Array{ElType, length(dims)}, dims)
+end
+function Base.similar(::Broadcasted{ChunkStyle{0}}, ::Type{Bool}, dims)
```
felixcremer, line 49: Why do we have to special case Bool here?

felixcremer:
Looks good to me. Thanks.

src/broadcast.jl
sethaxen: > Why do we have to special case Bool here?

Just so that it returns a `BitArray` like it does in base instead of an `Array`:

```julia
julia> typeof(fill(false) * true)
BitArray{0}
```

Base has both of these overloads: https://github.com/JuliaLang/julia/blob/01a2eadb0474c395845e66ed3382b52e0c1f1b8f/base/broadcast.jl#L228-L231

meggart:
Thanks a lot. Should I make a patch release?

asinghvi17:
Looks like this PR also updated the version number so may as well

sethaxen:
> Thanks a lot. Should I make a patch release?

That would be great!

PR #293, opened 2026-02-20
Fix iteration for 0-dimensional disk arrays

### LINKED ISSUE #289: Iteration fails for 0-dimensional disk-arrays
```julia

julia> a = UnchunkedDiskArray(fill(rand()));

julia> iterate(fill(1.0)) # works for Arrays
(1.0, 2)

julia> iterate(a) # fails for DiskArrays
ERROR: MethodError: no method matching OffsetArrays.OffsetArray(::Float64, ::CartesianIndices{0, Tuple{}})
The type `OffsetArrays.OffsetArray` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  OffsetArrays.OffsetArray(::AbstractArray, ::Any...; kw...)
   @ OffsetArrays ~/.julia/packages/OffsetArrays/b7qpm/src/OffsetArrays.jl:207
  OffsetArrays.OffsetArray(::OffsetArrays.OffsetArray, ::Tuple{Vararg{Int64}}; checkoverflow)
   @ OffsetArrays ~/.julia/packages/OffsetArrays/b7qpm/src/OffsetArrays.jl:179
  OffsetArrays.OffsetArray(::OffsetArrays.OffsetArray, ::Tuple{Integer, Vararg{Integer}}; kw...)
   @ OffsetArrays ~/.julia/packages/OffsetArrays/b7qpm/src/OffsetArrays.jl:186
  ...

Stacktrace:
 [1] _iterate_disk(a::UnchunkedDiskArray{Float64, 0, Array{Float64, 0}})
   @ DiskArrays ~/projects/DiskArrays.jl/src/iterator.jl:104
 [2] iterate(a::UnchunkedDiskArray{Float64, 0, Array{Float64, 0}})
   @ DiskArrays ~/projects/DiskArrays.jl/src/iterator.jl:112
 [3] top-level scope
   @ REPL[29]:1
```

sethaxen:
Fixes #289

### DIFF
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -61,11 +61,9 @@ end
 # Implementaion macros
 
 # Nested iteration over chunks
-@noinline function _iterate_disk(
-    a::AbstractArray{T}, i::I
-) where {T,I<:Tuple{A,B,C}} where {A,B,C}
+@noinline function _iterate_disk(a::AbstractArray{T}, i) where {T}
     # Split the data, block indices and state from the iterator
-    currentdata::A, blockinds::B, state::C = i
+    currentdata, blockinds, state = i
     # And split the block stat into the chunk iterator and inner indices
     (chunkstate, innerstate) = state
     # Need to check now as state will be updated
@@ -84,10 +82,10 @@ end
             newchunk = newinnerstate.itr.indices
             # Get a new chunk of data
             newdata = OffsetArray(a[newchunk...], newinnerstate.itr)
-            return newdata[i]::T, (newdata, blockinds, newstate)::I
+            return newdata[i]::T, (newdata, blockinds, newstate)
         else
             # Current chunk still has values left to iterate over
-            return currentdata[i]::T, (currentdata, blockinds, newstate)::I
+            return currentdata[i]::T, (currentdata, blockinds, newstate)
         end
     end
 end
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -61,7 +61,7 @@ end
 # Implementaion macros
 
 # Nested iteration over chunks
-@noinline function _iterate_disk(a::AbstractArray{T}, i) where {T}
+@noinline function _iterate_disk(a::AbstractArray{T}, i::I) where {T, I}
     # Split the data, block indices and state from the iterator
     currentdata, blockinds, state = i
     # And split the block stat into the chunk iterator and inner indices
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -91,6 +91,8 @@ end
         end
     end
 end
+_iterate_disk(a::AbstractArray{<:Any,0}, i::Int=1) = i == 1 ? (a[], 2) : nothing
+
 @noinline function _iterate_disk(a)
     # Get the indices for each chunk of data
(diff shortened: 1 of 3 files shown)


src/iterator.jl
```diff
@@ -91,6 +91,8 @@ end
         end
     end
 end
+_iterate_disk(a::AbstractArray{<:Any,0}, i::Int=1) = i == 1 ? (a[], 2) : nothing
```
felixcremer: Why is the index restricted to  an `Int`?  
According to the docstring of iterate this is supposed to be an Integer and that also works for normal Arrays.

sethaxen: Ah, I see now it doesn't even require the state be an `Int` (that's just for `AbstractString`). `iterate(::Array, i)`'s [implementation](https://github.com/JuliaLang/julia/blob/5fe89b8ddc166260bfcd4a195b305aff0ccad686/base/array.jl#L901) has no type constraint on `i` and just uses it for bounds checking and indexing. Meanwhile, valid index types for `AbstractArray`s are `CartesianIndex` (implicitly not supported here due to the bounds check) or "non-boolean integer" (https://docs.julialang.org/en/v1/manual/arrays/#man-supported-index-types), so I think this would be more consistent with the Base method, as it will error whenever the Base method would error (namely, when a `CartesianIndex` or `Bool` is provided):

```suggestion
_iterate_disk(a::AbstractArray{<:Any,0}, i=1) = i == 1 ? (@inbounds a[i], 2) : nothing
```

felixcremer: There is one equal sign to much in your function definition. Apart from that this looks good.

sethaxen: Ah, seems removing the Int specialization causes ambiguity with the other method. I could also remove that specialization or add it back to this method.

sethaxen: @felixcremer which do you prefer

meggart: In case you are still on this I would suggest to remove the other specialization as well...

sethaxen: > In case you are still on this I would suggest to remove the other specialization as well...

Done!

sethaxen:
@felixcremer @meggart Would it be possible to review this? The original issue is interfering with typical usage of one of my packages.

src/iterator.jl
```diff
@@ -61,11 +61,9 @@ end
 # Implementaion macros
 
 # Nested iteration over chunks
-@noinline function _iterate_disk(
-    a::AbstractArray{T}, i::I
-) where {T,I<:Tuple{A,B,C}} where {A,B,C}
+@noinline function _iterate_disk(a::AbstractArray{T}, i) where {T}
```
asinghvi17:
```suggestion
@noinline function _iterate_disk(a::AbstractArray{T}, i::I) where {T, I}
```

Could be worth it to still specialize especially with the noinline.

asinghvi17:
Looks good to me from a quick read!

felixcremer:
Thanks.

sethaxen:
> Thanks.

You're welcome! The PR also updated the patch version. Would you mind triggering a release as well?

PR #296, opened 2026-05-07
Call `checkbounds` also when using non-scalar indexing

lupemba:
This will also call `checkbounds` when using get index with slices, ranges etc. Previous the `checkbounds` where only called when doing scalar indexing. 

I think the missing `checkbounds` is a bug but there is a risk that some package might have used it as a feature.


asinghvi17:
Thanks, this looks like an improvement to me!  But should this get a `@propagate_inbounds` and `@boundscheck`?

lupemba:
@asinghvi17
Thanks for the feedback. I have added `@propagate_inbounds` and `@boundscheck` now. I have also moved the bounds check logic to the `Base.getindex` method. 

 Is this use of `@propagate_inbounds`  correct? I have never used these macros before.

rafaqz (approved):

Issue #273, opened 2025-08-27
threaded algorithms?


bjarthur:
i've just written a threaded version of a DiskArray-aware `unique` method:

```
uniqueT(v::AbstractDiskArray) = uniqueT(identity, v)
function uniqueT(f, v::AbstractDiskArray)
    u = Vector{Vector{eltype(v)}}(undef, length(eachchunk(v)))
    Threads.@threads :greedy for (i,c) in enumerate(eachchunk(v))
        u[i] = unique(f, v[c...])
    end
    reduce(u) do acc, t
        unique!(f, append!(acc, t))
    end
end
```

for a Zarr of size (666, 766, 3233) with 6732 chunks it is 4-5x faster using ~64 threads.  probably limited by the i/o of the mounted fileshare.

am i correct in thinking that to include such a method in DiskArrays.jl we'd need to know whether the storage format that backs the DiskArray is thread safe when read?  maybe with a type parameter?

has including threaded algorithms been discussed before?

happy to submit a PR if interested.

rafaqz:
Sounds good.

I think we will need a threading trait, that would be no threading by default, then backends can opt in.

We also probably need a way to turn it off globally like how `allowscalar` works with a `Ref{Bool}` const.

PR #308, opened 2026-09-15
Add a hook to switch compute backends

meggart:
As part of the discussion in #273 we discussed the necessity for optional compute backends. This is a draft implementation that let's users switch backends using Preferences.jl. It is also possible to set the backend to "dynamic", which let's you switch backends at runtime but involves some dynamic dispatch each time a computation is triggered (a cost which should be negligible, given we are concerned with operations that involve IO). A first backend would be DiskArrayEngine, the PR is here: https://github.com/meggart/DiskArrayEngine.jl/pull/69


src/compute.jl
```diff
@@ -0,0 +1,44 @@
...
+get_backend(b::ComputeBackend) = b
+
+# Default Backend, simple nonthreaded, non-distributed computations
+struct DefaultBackend <: ComputeBackend end
+
+# Wrapper type that lets users switch backends dynamically as the program runs
+# TODO add user interface for dynamic backend switching
+mutable struct DynamicBackend <: ComputeBackend
+    current_backend::ComputeBackend
+end
```
asinghvi17, line 15: To make this statically knowable, we can state that DiskArrays knows every backend that you can put into the DynamicBackend - and then others you have to pass through via `withbackend(array, SomeBackend())`.

asinghvi17: Then this can become a sumtype which cuts the dynamic dispatch into an if statement

src/compute.jl
```diff
@@ -0,0 +1,44 @@
...
+
+function load_backend()
+    @static if backend == "default"
+        DefaultBackend()
+    elseif backend == "dynamic"
+        return DynamicBackend(DefaultBackend())
+    elseif backend == "DiskArrayEngine"
+        DAE = Base.get_extension(@__MODULE__, :DiskArrayEngineExt)
+        isnothing(DAE) && error("Please add DiskArrayEngine to your environment to set it as a backend")
+        return DAE.DiskArrayEngineBackend()
```
asinghvi17, line 38: Let's define `DiskArrayEngineBackend` in DiskArrays directly and then have a good set of method error hints

src/compute.jl
asinghvi17: To be clear, I am proposing that we have a new function called `withbackend` which will return some wrapper like a WithBackendDiskArray(array, backend).  Then at all entry points where we determine the backend in order to dispatch, we can have a recursive function that checks and goes back to eg. `parent(array)` and has a check that if `parent(array) === array` then it stops to avoid infinite recursion.

asinghvi17: The sumtype approach is the following

```
struct DynamicBackend
    backend::Union{DefaultBackend, DiskArrayEngineBackend}
end
```
as both these types are concrete this is completely type stable and will trim et cetera.  dispatch should then become an if statement via union splitting.  measure this.

meggart:
Replaced by #316

PR #310, opened 2026-09-17
Add RangeIndex: index with a list of ranges, read in grouped blocks

asinghvi17:
`RangeIndex(1:2, 6:8)` indexes a disk array with a list of ranges.

- It is an `AbstractVector{Int}` of the concatenated ranges, so wrappers such as DimensionalData.jl pass it through unchanged and can slice their lookups with it. Checked through `DimArray` and a lazy GeoTIFF `Raster` with no changes to either package.
- `process_index` groups the ranges into blocks: sorted by start, consecutive ranges are merged unless a whole chunk lies in the gap between them, which is the only I/O a split saves. With 4-row chunks, `a[RangeIndex(1:2, 6:8), 2]` is one `readblock!` of `1:8`; with 2-row chunks it is two.
- Slicing it with a range gives a `RangeIndex`, so reads through a `view` stay grouped. `setindex!` writes the same blocks.

Spike history: the first commit added a `Colon`-like index to map the path; it is removed in the second.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NPFXtV2iEJJjj5ZTfDXMRN


asinghvi17:
Two pre-existing bugs found on the way, neither involving this PR. Both reproduce on `main`:

```julia
using DiskArrays, DiskArrays.TestTypes
a = AccessCountDiskArray(reshape(collect(1:64), 8, 8); chunksize=(2, 8))
```

1. Batched `setindex!` with an `Integer` in another dimension:
   ```julia
   a[[1, 3, 7], 2] = 1:3   # MethodError: no method matching numind(::Int64)
   ```
   `setindex_disk_batch!` calls `is_sparse_index` on every temp index, and there is no method for integers. This PR adds `is_sparse_index(::Integer) = false`, since `RangeIndex` writes hit it too.

2. An empty range in a batched read:
   ```julia
   a[[1, 6], 1:0]   # ArgumentError: invalid Array dimensions
   ```
   The `ChunkRead` `process_index` for integer arrays starts `maxtempindex = -1` and never updates it for an empty index, so the temp array gets a negative size. Not fixed here.

src/rangeindex.jl
```diff
@@ -0,0 +1,77 @@
...
+
+Base.size(i::RangeIndex) = (sum(length, i.ranges; init=0),)
+function Base.getindex(i::RangeIndex, k::Int)
+    @boundscheck checkbounds(i, k)
+    for r in i.ranges
+        k <= length(r) && return r[k]
+        k -= length(r)
+    end
+    throw(BoundsError(i, k))
+end
```
meggart, line 29: One potential optimization here could be to store the cumulative offsets of the ranges in another array in RangeIndex. Then one could avoid the loop in getindex and replace it with a searchsorted.

src/rangeindex.jl
```diff
@@ -0,0 +1,77 @@
...
+    @boundscheck checkbounds(i, k)
+    ranges = UnitRange{Int}[]
+    offset = 0
+    for r in i.ranges
+        lo, hi = max(first(k), offset + 1), min(last(k), offset + length(r))
+        lo <= hi && push!(ranges, r[lo-offset]:r[hi-offset])
+        offset += length(r)
+    end
+    return RangeIndex(ranges)
+end
```
meggart, line 39: This optimization is good to have. Do indexing with `Colon` or calling `collect` run through this path as well?

src/rangeindex.jl
```diff
@@ -0,0 +1,77 @@
...
+    group_ranges(ranges, chunks::ChunkVector) => Vector{Vector{Int}}
+
+Group `ranges` (by position) so that each group is read as one block, with as few
+blocks as possible at no extra chunk I/O: consecutive ranges, sorted by start, are
+split only when a whole chunk lies in the gap between them.
+"""
+function group_ranges(ranges, chunks::ChunkVector)
+    groups = Vector{Int}[]
+    stop = 0 # last data index of the current group
+    for k in sortperm(ranges; by=first)
```
meggart: Ah I see, the ranges are not sorted by default. I this something we really need from the DimensionalData/DGGS side? It would make things a lot easier if during construction we could make sure that a RangeIndex is sorted and non-overlapping, but I do not know if it would still cover the most important use cases.

src/rangeindex.jl
```diff
@@ -0,0 +1,77 @@
...
+    end
+    return RangeIndex(ranges)
+end
+Base.show(io::IO, i::RangeIndex) = print(io, "RangeIndex(", join(i.ranges, ", "), ")")
+Base.show(io::IO, ::MIME"text/plain", i::RangeIndex) = show(io, i)
+
+# Always batch, grouping the ranges into blocks. `NoBatch` keeps the
+# `AbstractArray{<:Integer}` behaviour, a single read of the hull.
+_need_batch_index(::RangeIndex, chunks, ::Union{ChunkRead,SubRanges}) = true, tail(chunks)
+for S in (:ChunkRead, :SubRanges) # one method per strategy, to beat the `AbstractArray{<:Integer}` ones
```
meggart, line 56: I think the solution implemented here (based on accessing every chunk only once) exactly matched the model of the `ChunkRead` strategy. However, for backends using `SubRanges` strategy we should use a simpler strategy that just loops through the ranges provided by the `RangeIndex` and reads these, relying on the cache provided by libnetcdf etc to avoid repeated access of the same chunk on disk.

src/rangeindex.jl
asinghvi17: Yeah I think we should ensure that

asinghvi17:
Benchmarks at 133f48f: `RangeIndex` vs a `Vector{Int}` of the same indices. Disk arrays are `AccessCountDiskArray`s over in-memory data (so "time" is index processing and copying, no real I/O), except the last section, a 4000×4000 Float32 GeoTIFF with 256×256 tiles read lazily through Rasters. Minimum of ≥1 s of samples.

**Reads and elements read are identical in every case**, since a `RangeIndex` now batches exactly when its `collect` would; only the block shapes differ. Times:

| case | reads | RangeIndex | Vector{Int} |
|---|---|---|---|
| 1-d 10⁶, chunks 10⁴: 100 ranges of 500 every 1 000 (dense → hull) | 1 | 0.25 ms | 0.08 ms |
| 1-d: 100 ranges of 500 every 10 000 (sparse) | 100 | 0.09 ms | 0.58 ms |
| 1-d: 10 000 ranges of 5 every 100 | 100 | 0.98 ms | 1.32 ms |
| 1-d: 1 000 single indices every 1 000 | 100 | 0.61 ms | 0.59 ms |
| 2-d 4000², chunks 256²: `[r, 1:200]`, 20 ranges of 100 | 1 | 1.14 ms | 1.30 ms |
| 2-d `[r, 1:200]`, 20 ranges of 20 (sparse) | 13 / 15 | 0.30 ms | 0.35 ms |
| 2-d `[r, r]` sparse | 169 / 225 | 4.9 ms | 4.9 ms |
| 3-d 500×500×100, chunks 64×64×10: `[r, r, rz]` | 100 | 0.56 ms | 0.63 ms |
| 1-d unchunked (`SubRanges`): 10 000 ranges of 5 | 10 000 | 2.3 ms | 5.1 ms |
| 1-d 10⁷: 100 000 ranges of 5 | 1 000 | 14.2 ms | 17.4 ms |
| `DimArray` over the 1-d array, 10 000 ranges: `A[r]` | 100 | 1.9 ms | 1.4 ms |
| GeoTIFF `R[X(r), Y(1:200)]`, dense / sparse / 400 tiny | | 3.3 / 4.0 / 3.5 ms | 3.3 / 4.1 / 3.5 ms |

(The two sparse 2-d rows read 13 vs 15 and 169 vs 225 blocks because adjacent ranges join across chunk boundaries.)

Where the vector still wins: the dense hull read, by the cost of the `collect` in `to_indices` (~0.15 ms per 50 000 elements), and slicing a DimensionalData `Sampled` lookup, which calls `to_indices` on the lookup before the range underneath sees the `RangeIndex` (0.7 vs 0.02 ms for 10 000 ranges). Both are small next to a `readblock!`.

Why `to_indices` rather than an iterator (from the earlier review discussion): Base reaches an index through `getindex` inside a `view` (`Base.reindex`) and through `iterate` in `A[i]`, and even a fast iterator is ~10× slower than a gather through a `Vector{Int}` because the loop doesn't vectorize. Materializing for `DenseArray`s and ranges only gives vector speed where the copy happens while disk arrays and wrappers keep the structured index.

meggart (approved):

meggart:
Do we still want to try to support the n-dimensional case or do we leave this for later? I would suggest to leave it for now.

asinghvi17:
I think we can leave the n-dimensional case for later, yeah

PR #312, opened 2026-09-18
Add extensible chunkexists queries

### LINKED ISSUE #301: API function to query for whether a certain chunk exists or not
Zarr already has isinitialized so this is basically that.  Also needed for regridding usecases so we can skip chunks that never exist.

asinghvi17:
Adds `chunkexists` so backends with optional chunk storage can report whether a chunk is stored. The scalar fallback returns `true`; backends can override it to report absent chunks, which may still read as fill values.

Accepts one-based integer chunk coordinates, tuples, `CartesianIndex`, and `ChunkIndex`. The iterable form calls the scalar method by default, preserves array input shape, and can be specialized for batched storage queries. Callers supply valid chunk coordinates; the fallback performs no bounds checks or data reads.

This is mainly meant for multi-file datasets like Zarr (fixes #301).  But I guess there could be other formats which cause empty chunks to somehow vanish (outside of compression).  I'm not sure whether we want to apply this to the concat disk array, though I suspect we should.

Validation: full `Pkg.test()` suite passed, including 32 new tests for defaults, scalar overrides, iterators, empty batches, index adapters, and batch overrides, plus Aqua checks.

Made with 🤖 GPT-6-Astra (codex)


test/chunkexists.jl
```diff
@@ -0,0 +1,75 @@
+struct ChunkExistenceArray{N} <: AbstractDiskArray{Int,N}
```
meggart: For testing the new functionality, it would be great to use the new `ConcatDiskArray` that contains `MissingTile` as missing chunks.  We should definitely implement `chunkexists` for this type and maybe when we have it, we do not need to define an extra type for testing.

meggart (changes requested):

asinghvi17:
Addressed the review in 2227d22: `chunkexists` is now implemented for `ConcatDiskArray` (`src/cat.jl`). It locates the tile holding the chunk, returns `false` for a `MissingTile`, and otherwise forwards the query to the parent tile's own chunk. The tests in `test/chunkexists.jl` now use `ConcatDiskArray` grids with `MissingTile` (2D mosaic, stacking along a new dimension, and nested concat) instead of the ad-hoc test types. Full test suite passes locally.

meggart (approved):

meggart:
> in [2227d22](https://github.com/JuliaIO/DiskArrays.jl/commit/2227d2251f231d88ba59fadc5a1ccacdb2cb3430): `chunkexists` is now implemented for `ConcatDiskArray` (`src/cat.jl`). It locates the tile holding the chunk, returns `false` for a `MissingTile`, and otherwise forwards the query to the parent tile's own chunk.

Thank you. Now I realize that the implementation for ConcatDiskArray was more complex than I assumed because of there is not 1:1 relationship between tiles and chunks. Glad you figured it out anyway and it is good to have this possibility of early escape for mosaics with missing tiles which might be lazy-loaded in many cases.

meggart (approved):

